PR #39, opened 2015-01-11
Empty translations replacement

mod3st:
Added the ability to return the original models attribute values for empty translations.
Using example: $model->lang('pl')->skipEmptyTranslations()


daftspunk:
I think this should be enabled by default and it should be called "fallback".

So `$model->lang('pl')->noFallback()` would mean empty values are empty and do not fall back to the default language.

mod3st:
Hey, just added proposed changes:
- "fallback" calls
- fallbacks as default option

mod3st:
Hey, what about this change?

daftspunk:
Sorry - was busy with RC, finally found the time. 

Thanks for this!

daftspunk:
Note I have renamed `noFallbackTranslations` to  `noFallbackLocale`

Issue #18, opened 2014-08-23
Translations when creating Models


alxy:
Currently, translations do not work in Popups/Modals. Screenshot: https://www.dropbox.com/s/13lo069zldzkmqx/Screenshot%202014-08-23%2015.36.34.png?dl=0

The POST-data looks correct, it includes the translation strings. Also, the TranslatableModel::setTranslateAttribute() function is called accordingly. However, in the syncTranslatableAttributes()-method the $translateableAttributes array is empty (empty array).
My guess is, the model / model behaviour is re -instanciated somewhere in the application. It may be related to the way october saves related models, but I couldnt investigate this any further.
I could somehow verify my guess by specifying a default value on the array:
```
protected $translatableAttributes = ['test'];
```
In this case, it really contained this value in the syncTranslatableAttributes()-method.

Hope this helps a bit. to identify the error.

daftspunk:
Noted this only occurs during creation. Subsequent updates are fine.

PR #50, opened 2015-03-12
Fixes bug saving data from modals

ariponce:
Fixes https://github.com/rainlab/translate-plugin/issues/18 where translation were not saved when the form was called from a modal.


ariponce:
It still doesn't work for me, saved values are always empty strings when using a relation behavior form, even on updates.

Kunilt:
+1 still doesn't work. 
Trying to save (already created model) from pop up and translations are not saved to database.

daftspunk:
Please submit another issue with replication instructions.

Issue #42, opened 2015-01-23
Incorrect locale while going from /locale/link to /link


is-blackhole:
I have found this issue: 
I have 2 langs configured, Russian as default and English.
If you go for example to site.com/en/about , then go directly to site.com/about locale will not be changed to default. Thats because of this code in routes.php:

``` php
    $locale = Request::segment(1);
    if ($translator->setLocale($locale)) {
        Route::group(['prefix' => $locale], function() use ($locale) {
            Route::any('{slug}', 'Cms\Classes\Controller@run')->where('slug', '(.*)?');
        });
        Route::any($locale, 'Cms\Classes\Controller@run');
    }
```

setLocale will store locale in session, then later in Plugin.php locale will be loaded from session for translation. But going to site.com/about will not change locale in session since no locale specified in url and setLocale will fail.
This can be fixed with adding else:

``` php
    $locale = Request::segment(1);
    if ($translator->setLocale($locale)) {

        Route::group(['prefix' => $locale], function() use ($locale) {
            Route::any('{slug}', 'Cms\Classes\Controller@run')->where('slug', '(.*)?');
        });

        Route::any($locale, 'Cms\Classes\Controller@run');
    } else $translator->setLocale($translator->getDefaultLocale());
```

SaifurRahmanMohsin:
This isn't an issue but rather a feature. The way the pages are supposed to work are that they remember the language setting across pages. In fact, on my website if I switch the locale to a different language then it will show the page in the other language without adding the /LANG_CODE/about and instead in the same route page i.e. /about

oliverpool:
Easier fix in the routes.php file:
```
if ($translator->setLocale($locale, false)) {
```
 Set to false the second parameter of `setLocale` to not store in the session (see [the code](https://github.com/rainlab/translate-plugin/blob/master/classes/Translator.php#L55) )

LukeTowers:
Closing as it has been over a month since any activity on this occurred.

PR #57, opened 2015-04-11
Add local to session

romanov-acc:
Add this changes please :)


oliverpool:
Hi,
could you describe the goal of this PR?

It is something similar to #42 ??

PR #59, opened 2015-04-19
Change check for translation in getTranslateAttribute

niclasleonbock:
Before, the fallback was useless. It was never called because `$this->translatableAttributes[$locale][$key]` is never unset but just empty.


daftspunk:
Great, thanks!

Issue #11, opened 2014-08-09
Default root (/) language


vizo:
In documentation is written:
- http://website/ru/ will display the site in Russian
- http://website/fr/ will display the site in French
- http://website/ will display the site in the default language or the user's chosen language.

I think, if you have multilingual site, it is better that you always redirected to /language when you come to root. Joomla have the same option. Why?
- people often don't look at url, they just make bookmark, share, etc., so if you change default language later, content could be gone on same url, which is very bad.
- for search engines content is duplicated on two urls: / and /language, which is also bad

Is it possible to make this as an option, that you are always redirected?

So when user come to root, it checks:
1. if user have language in session, redirect to user's language
2. if browser sends language in request header, redirect to browser's language
3. redirect to default language

Thanks in advance,
vizo

daftspunk:
Fixed by #72 and should be available soon.

PR #72, opened 2015-05-12
Added option to always prefix routes with language code

justin-lau:
This PR adds an option in the backend setting to always prefix routes with language code.

Pretty much described in #11, if it is turned on, when a CMS route (not backend, not routes defined in other plugins such as the `debugbar`) is accessed without the locale prefix, it will check against the browser session for the language to use, or fallback to the default language, and returns a 302 redirect to a URL with the prefix. The locale picker will also issue 302 redirects when the languages is changed.

Language check from request header isn't implemented in this PR, as the language code from the browser may not match the user defined language code.

### DIFF
--- a/Plugin.php
+++ b/Plugin.php
@@ -103,8 +103,9 @@ class Plugin extends PluginBase
     public function registerPermissions()
     {
         return [
-            'rainlab.translate.manage_locales'  => ['tab' => 'Translation', 'label' => 'rainlab.translate::lang.plugin.manage_locales'],
-            'rainlab.translate.manage_messages' => ['tab' => 'Translation', 'label' => 'rainlab.translate::lang.plugin.manage_messages']
+            'rainlab.translate.manage_locales'     => ['tab' => 'Translation', 'label' => 'rainlab.translate::lang.plugin.manage_locales'],
+            'rainlab.translate.manage_messages'    => ['tab' => 'Translation', 'label' => 'rainlab.translate::lang.plugin.manage_messages'],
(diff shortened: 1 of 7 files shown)


justin-lau:
Rebased on version 1.0.11 9df390c817475c7c2457d68e217f7c5759ded1b2

lowerends:
Very useful pull request, thanks.

justin-lau:
I have added an extra preference to set the default frontend language to be redirected to. It behaves differently than the default language set in the "Manage Languages" tab, where changing default language affect model attribute translations.

keiosweb:
@daftspunk , do you plan to merge this?

shrpne:
+1 I need this option

Flo-Schield-Bobby:
+1!

daftspunk:
This will be merged soon, just need to find the time to test it. Thanks for your patience.

oryxgd:
+1

daftspunk:
This has been partially merged and is due to become available in the next major release. It seems a better idea to include the redirection logic as part of the localePicker component, rather than a global preference. This allows for greater control and is less magical with fewer assumptions. It means a localePicker component should be included on every page/layout that requires the redirection to occur.

See the `forceUrl` property in the `RainLab\Translate\Components\LocalePicker` component:

```
'forceUrl' => [
    'title'       => 'Force URL schema',
    'description' => 'Always prefix the URL with a language code.',
    'default'     => 0,
    'type'        => 'checkbox'
],
```

And usage on page/layout:

```
[localePicker]
forceUrl = 1
```

Thanks @justin-lau for your work on this.

PR #83, opened 2015-07-19
Fixed unicode message not translatable

### LINKED ISSUE #82: Messages in unicode don't get stored in the database correctly.
When translatable messages are scanned, each message is saved to the database along with its `code` generated by `Message::makeMessageCode()`, which uses Laravel's `Str::slug()` that tries to convert non-ascii characters into ascii. Unconvertible characters are simply removed. Therefore messages without ascii characters are stored with an empty `code` value in the database, rendering unicode messages untranslatable.

I assume `Str::slug()` is used to quickly generate unique keys for the message entries. If this is the case, maybe replacing it with a fast hashing algorithm will yield better result and also allow unicode messages to be translated?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/24572656-messages-in-unicode-don-t-get-stored-in-the-database-correctly?utm_campaign=plugin&utm_content=tracker%2F4122124&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F4122124&utm_medium=issues&utm_source=github).
</bountysource-plugin>

justin-lau:
This PR fixes #82. I have created it based on an assumption that the message codes are merely used as a key of the hash table of messages. If I have overlooked anything, hashing may not be the way to solve the unicode message issue. Anyway, I need unicode messages to be translatable in my current project and I have tested and confirmed it working great with unicode messages as well as the English based ones.

The message code is replaced by a hash value using [MurmurHash 3](https://github.com/lastguest/murmurhash-php), the [fastest non-cryptographic hashing algorithm](https://github.com/Cyan4973/xxHash#benchmarks) I can find without the need of an extension, yet has a fairly well collision avoidance. (Extended reading: [Hash functions: An empirical comparison](http://www.strchr.com/hash_functions))

This PR changes the way the message code is generated, therefore **it breaks the ability to lookup previously stored messages**.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+/vendor
+composer.lock
+.DS_Store
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,11 @@
+{
+    "require-dev": {
+        "keios/oc-plugin-testsuite": "dev-bugfix"
+    },
+    "repositories": [
+        {
+            "type": "git",
+            "url": "https://github.com/justin-lau/oc-plugin-testsuite"
+        }
+    ]
+}
--- a/composer.json
+++ b/composer.json
@@ -1,11 +1,5 @@
 {
     "require-dev": {
-        "keios/oc-plugin-testsuite": "dev-bugfix"
-    },
-    "repositories": [
-        {
-            "type": "git",
-            "url": "https://github.com/justin-lau/oc-plugin-testsuite"
-        }
-    ]
+        "keios/oc-plugin-testsuite": "dev-master"
+    }
 }
--- a/models/Message.php
+++ b/models/Message.php
@@ -212,6 +212,25 @@ class Message extends Model
      */
     protected static function makeMessageCode($messageId)
     {
+        $separator = '.';
+
+        // Convert all dashes/underscores into separator
+        $flip = $separator == '-' ? '_' : '-';
+        $messageId = preg_replace('!['.preg_quote($flip).']+!u', $separator, $messageId);
+        // Remove all characters that are not the separator, letters, numbers, or whitespace.
+        $messageId = preg_replace('![^'.preg_quote($separator).'\pL\pN\s]+!u', '', mb_strtolower($messageId));
+        // Replace all separator characters and whitespace by a single separator
+        $messageId = preg_replace('!['.preg_quote($separator).'\s]+!u', $separator, $messageId);
+
+        return Str::limit(trim($messageId, $separator), 250);
+    }
+
+    /**
+     * @deprecated
+     * @TODO Remove this function in the next major version.
+     */
+    protected static function makeDeprecatedMessageCode($messageId)
+    {
         return Str::limit(strtolower(Str::slug($messageId, '.')), 250);
     }
 
--- a/models/Message.php
+++ b/models/Message.php
@@ -102,11 +102,24 @@ class Message extends Model
         ]);
 
         /*
+         * Copy deprecated message data over if exists.
+         *
+         * TODO: Remove this sinppet in the next major version.
+         */
+        if (!$item->exists) {
+            $deprecatedItem = static::whereCode(self::makeDeprecatedMessageCode($messageId))->first();
+
+            if ($deprecatedItem) {
+                $item->message_data = $deprecatedItem->message_data;
+            }
+        }
+
+        /*
          * Create a default entry
          */
         if (!$item->exists) {
             $data = [static::DEFAULT_LOCALE => $messageId];
-            $item->message_data = $data;
+            $item->message_data = $item->message_data ?: $data;
             $item->save();
         }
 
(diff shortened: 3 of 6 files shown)


daftspunk:
Hmm, I'm not sure about this, we just need an mbslug function like we have in the [JS lib input.preset](https://github.com/octobercms/october/blob/master/modules/system/assets/ui/js/input.preset.js).

justin-lau:
The slug function in the JS lib is similar to Laravel's `Str::slug()` in that it tries to convert unicode characters into similar ascii characters. While it works for syllable based languages, it trims off the non-convertible characters from other languages.

Here is an example using the JS lib's function:

```js
slugify('Hello World!'); // works for English
// -> "hello-world"

slugify('Hello Piñata!'); // and Spanish
// -> "hello-pinata"

slugify('Hello 皮納塔!'); // but not for Chinese
// -> "hello"

slugify('Hello ピニャータ!'); // nor Japanese
// -> "hello"

slugify('Hello 피냐타!'); // nor Korean
// -> "hello"
```

Slug generating functions are great for URL generation because it makes them human readable, but it cannot generate 1:1 mapping as long as unicode characters is concerned. I think hash function generates unique keys better and faster, unless there is an absolute necessity to make the code readable.

daftspunk:
I see the problem now. We'll need to come up with a way to make this backwards compatible by converting existing translations, or leave it until the next major version as a breaking change.

justin-lau:
**1. Perhaps it could be optional by configuration?**

Since this plugin is intended for handling translations, unicode messages should be supported by default.

I am thinking of the way EmberJS team does to their Javascript framework. They make new features *temporarily* backward compatible, and warn the developer of deprecations. In the next major release they simply remove the backward compatibilities and warnings. We may be able to do the same here, and it would be more developer friendly than using configuration.

**2. Is this hashing function superior to something inbuilt, like md5?**

As far as I know, PHP's built in hashing functions are all for cryptographic uses. They serve a different purpose than non-cryptographic hash functions like MurmurHash, and tend to run slower. MurmurHash3 in particular, runs faster and produces fewer collisions compared to most other functions.

**3. What about simply using strtolower(str_replace(' ', '-', ...)) ?**

I've actually considered this, but I'm still missing the point to make the message code human readable. As long as the message code simply serves as mapping keys and human readability is not a concern, hashing function does the job faster and better.

Also, `strtolower(str_replace(' ', '-', ...))` does not generate 1:1 mapping keys. I am not sure whether that is the way to go for translating messages. For example, `"Hello World"` may have different translation than `"hello world"` (letter casing, different meaning in context, etc.), but they both have the same code `"hello-world"`.

daftspunk:
To give you some more context, the language system is very similar to the one found in the back-end where you see a language key `account.sign_out` with a value of **Sign out**. The current behavior used is an attempt at automatically generating the key name from the value itself, this is considered a "relaxed" way to approach translation. It is easier and more fluid, but conflicts can occur like in the Hello World example above.

A planned feature we have for this plugin, in future, will be a way to store seed translations in **/themes/..../meta/lang.yaml** or some other location. This enables the ability to use key based translations instead, i.e `{{ 'account.signout'|_ }}` and would resolve the issues with conflicts. This is considered the "strict" approach to translation. It gives complete granular control, and would also allow themes to come pre-translated.

In summary we need human readable keys to exist because they are nicer to humans, and coexist better with both relaxed and strict approaches.

justin-lau:
Now that I know about the planned feature, I understand why readable key is necessary. I think [this slug function](https://gist.github.com/sgmurphy/3098978) is the one we need.

In short, it does this:

```php
$options = [
    'delimiter' => '.',
    'limit' => 250,
];

url_slug('Hello Pinata!', $options);    // => "hello.pinata!"
url_slug('Hello Piñata!', $options);    // => "hello.piñata"
url_slug('Hello 皮納塔!', $options);     // => "hello.皮納塔"
url_slug('Hello ピニャータ!', $options);  // => "hello.ピニャータ"
url_slug('Hello 피냐타!', $options);      // => "hello.피냐타"
```

justin-lau:
I have grabbed Laravel's `Str::slug()` code, removed the line it translates and removes the non-ascii characters, and it becomes the new `makeMessageCode()`.

In order to make it backward compatible, I keep the old function but renamed it to `makeDeprecatedMessageCode()` and adds logic to check against the database whenever a new message is detected in the system. The new message will have any existing old message data copied over.

I have also added unit and functional tests to verify the changes.

daftspunk:
This looks great, I will merge it soon.

justin-lau:
Nice! I have just updated the `oc-plugin-testsuite` version to point it back to the master branch, as some PRs I submitted have been merged.

daftspunk:
Thanks for this!

daftspunk:
Just FYI - I have updated this code further to use a newly created `PluginTestCase` added to October. You may wish to pull this in to your project so it works with the latest plugin code (or wait for the next build!)

See commit 0c8e813029ccf0c86f40e2dce444c1b2234adcbd

justin-lau:
Cool :+1:

PR #104, opened 2015-11-17
Minor fix for trans method

estudiogenius:
It makes it easier to use the method inside Controllers and Components


daftspunk:
LGTM!

PR #125, opened 2016-04-22
fixed translation not showing up on load

targetblank:
Textarea value is handled at inner HTML. Translated value should be moved from attribute "value" to inner HTML.


side-by-side:
I can confirm this bug too and correction suggested by @targetblank solve it.

But please merge my corrected pull request #137, not this one. In this version there is missing class that leads to wrong language selector view.

Ping @daftspunk

daftspunk:
Fixed by c8e60b47e93a8b7f55b305d88471f61f936f284c

Thanks!

PR #134, opened 2016-06-02
Fixed the flush message display

gergo85:
Before:
before

After:
after

The layout code is same, but the margin top of flush message is missing. I don't understand the reason.


daftspunk:
The "Before" is preferred over "After".

PR #163, opened 2016-08-17
Fire event on setLocale

rounce:
Justification. I need to remember the user's locale in the database to send him email or sms in his locale (for example from cron).


daftspunk:
Try this Laravel event, triggered by App::setLocale()

```
Event::listen('locale.changed', function($locale) { ... });
```

If you need it to only fire on the front end use an if statement with `App::runningInBackend()`

rounce:
Thanks. It works.

PR #176, opened 2016-10-12
(fix #174) change model_id field type to int

### LINKED ISSUE #174: invalid DB query
When using PostgreSql with latest translate-plugin and blog there is an exception thrown

`SQLSTATE[42883]: Undefined function: 7 ERROR: operator does not exist: integer = character varying LINE 1: ...b_translate_indexes" on "rainlab_blog_posts"."id" = "rainlab... ^ HINT: No operator matches the given name and argument type(s). You might need to add explicit type casts. (SQL: select "rainlab_blog_posts".* from "rainlab_blog_posts" left join "rainlab_translate_indexes" on "rainlab_blog_posts"."id" = "rainlab_translate_indexes"."model_id" and "rainlab_translate_indexes"."model_type" = RainLab\Blog\Models\Post and "rainlab_translate_indexes"."locale" = en where ("rainlab_blog_posts"."slug" = first-blog-post or ("rainlab_translate_indexes"."item" = slug and "rainlab_translate_indexes"."value" = first-blog-post)) and "published" is not null and "published" = 1 and "published_at" is not null and "published_at" < 2016-10-11 14:29:10 limit 1)
`

Please check **plugins/rainlab/translate/behaviors/TranslatableModel.php** L55

`->on($this->model->getQualifiedKeyName(), '=', 'rainlab_translate_indexes.model_id')`

Basically you are joining integer on varchar field which is not allowed without type casting.

webeks:


daftspunk:
`model_id` can be a GUID so it should remain as a string. We probably need to cast a value to `TEXT` somewhere...

PR #185, opened 2016-11-03
Update EventRegistry.php

mbrode:
missing &&


daftspunk:
Thanks, fixed already by #184. I've got to stop pushing changes while burning the midnight oil 😄

PR #199, opened 2016-12-17
Support maxlength attribute

mariavilaro:
Support maxlength attribute in text fields in the same way octobercms does


mariavilaro:
Added handler for pages.menuitem.resolveitem that works with translated urls even when there is no locale in the current url (for example when loading a sitemap.xml).

mariavilaro:
I'm sorry maybe I'd had to make two branches for my changes, one for every commit? I'm still learning github, sorry!

LukeTowers:
Nope, one branch is fine, we just need to find the time to review it. Thanks for your submission!

LukeTowers:
Do you mind explaining why this feature is desired / needed? Not against it, but more information on why it should be included means that we can process this faster.

mariavilaro:
maxlength attribute, I think this one is obvious: so you can use the maxlength attribute in backend forms

Handler for pages.menuitem.resolveitem: I need it to be able to generate multilanguage sitemaps, you can see why in my pull request in the sitemap plugin https://github.com/rainlab/sitemap-plugin/pull/43

PR #210, opened 2017-01-08
Set value to an empty string if empty

H6:
If this is not done, laravels escaping helper function `e` will through an error

> "htmlentities() expects parameter 1 to be string, array given" on line 469 of /vendor/laravel/framework/src/Illuminate/Support/helpers.php


aurukas:
should be merged. Simple but neccessary fix

LukeTowers:
@aurukas Please provide replication instructions for the issue that you are encountering where this would fix it

aurukas:
this was already replicated by @CptMeatball in the linked issue Repeater field error

fdeneux:
Fixed it for me.

LukeTowers:
Verified problem and solution, thanks for the fix @H6!

Issue #227, opened 2017-02-22
LocalePicker doesn't translate slugs on CMSPage


CptMeatball:
The JS implementation of the LocalePicker doesn't work flawlessly when using multiple multi language `slugs`.

### Steps to reproduce
Languages (2): NL (default) and DE
CMS URL: `/staticPageName/:category/:slug`
Custom Component that uses both `category` and `slug` parameters.

#### Actions
**Current URL:** `www.domain.tld/de/pagename_de/category_de/slug_de`
-> Switch language through `localePicker` component. 
In this case the URL becomes: `www.domain.tld/nl/pagename_nl/category_de/slug_de`. It fails to set the slugs to the new language.

CMS now tries to query the wrong language in the `transWhere` function of the model. As the default language is selected, it queries for the **untranslated** slugs.

Alternatively, if the non-default language is selected, it queries for the **translated** slugs, but in this case it has a fallback to the default language and the query is successful.

### Expected Behavior
LocalePicker components also translates the `slugs` when switching language.

wetgi:
I can confirm this

mariavilaro:
I'm having the same problem.

roulendz:
the same for me, I am using slug to open translated model, but I get Record not found!
I have custom component with slug and title.
For example
```
/lv/slug //works well
/ru/slug_in_ru // shows Record not found
```

mariavilaro:
I found a workaround for this, but maybe having an indicator that the language has just changed would be better than using the session:

    public function onRun()
    {
        $slug = $this->param('slug');

        if (Session::has('lastLocale') && Session::get('lastLocale') != $this->page['activeLocale']) {
            $news = NewsModel::transWhere('slug', $slug, Session::get('lastLocale'))->first();
            Session::put('lastLocale', $this->page['activeLocale']);
            if ($news) {
                $page = Page::loadCached(Theme::getActiveTheme(), 'news-detail');
                $page->rewriteTranslatablePageUrl();
                $url = Translator::instance()->getPathInLocale(str_replace(':slug', $news->slug, $page->url), $this->page['activeLocale']);
                return Redirect::to($url);
            }
        }

        $news = NewsModel::transWhere('slug', $slug)->first();
        if (!$news) {
            $this->setStatusCode(404);
            return $this->controller->run('404');
        }
        $this->page['news'] = $news;
        Session::put('lastLocale', $this->page['activeLocale']);
    }

LukeTowers:
[...]

Issue #167, opened 2016-09-09
Provide an API for translating slugs via onSwitchLocale


thomasilmer:
**I was told to create a new issue for this, this is a copy of my post in Translate page's URL issue #25.**

Hello @daftspunk , I have been testing the new feature and so far it's great! 

I just have one small issue with implementing the translatable slugs. Let's say I have a page with this signature:

```
url = "/projects/:slug"

[viewBag]
localeUrl[nl] = "/projecten/:slug"
localeUrl[de] = "/projekte/:slug"
```

In which ":slug" is different for every language, e.g.:
en => house
nl => huis
de => haus

Currently we have a custom plugin which has a backend menu for adding projects (Basically adds a Project model with some properties), which has a translatable index attribute "slug". In the onRun() function of "/projects/:slug" we check the database for a Project model which has the current slug as attribute. The page is then populated with properties of the Project model.

This works fine, untill you use the localePicker, because only the "projects/projecten/projekte" part of the url is switched. The ":slug" part stays the same as before the locale switch. This can result in people being redirected to "projects/huis" for example. This URL (rightfully) gives a 404. 

I tried looking at how RainLab.Pages plugin tries to solve this, but I haven't found the same functionality there.

What is a nice solution to translate the slug when the onSwitchLocale is called?

I also have a second issue which is somewhat similar. Let's say I don't use the language prefixes /en/, /nl/ or /de/, but instead use domains to determine the locale:

example.com => english
example.nl => dutch
example.de => german

Currently we set the session locale based on the domain and it works. Except again, when we switch language with the localePicker, the domain won't change with it. Currently we have a temporary fix by redirecting to the domain homepage (not using the localePicker), but it would be a lot nicer to redirect to the current page in a different language.

Is there an event for onSwitchLocale? Because then we could probably set the domain after onSwitchLocale ourselves, but currently I don't know how to fix this.

Sorry if this is in the wrong topic, but I felt like this is partly in the scope of the new translatableIndexes.

jcardell:
+1

Borgoroth:
+1

Fl0Cri:
+1 currently it's confusing to see that urls for static pages are translatable, but not slugs in blog posts (as mentioned by @thomasilmer, trying to translate a slug would result in a buggy situation where changing the language may lead to a 404)

roulendz:
Is there fix for this, I also have the same problem. 
Switching language the slug does not get translated. 

[...]

PR #244, opened 2017-04-02
Translate url params

mariavilaro:
I'm trying to solve this issue https://github.com/rainlab/translate-plugin/issues/227

What I did is to fire an event, so each model can register and decide how to translate their parameters. For example, for my news Model I did this. In Plugin.php:

        Event::listen('translate.params.translateParams', function($page, $params, $old_locale, $new_locale) {
            if ($page == 'news-detail') {
                return NewsModel::translateParams($params, $old_locale, $new_locale);
            }
        });

And in the model News.php:

    public static function translateParams($params, $old_locale, $new_locale) {
        $newParams = $params;
        foreach ($params as $paramName => $paramValue) {
            $news = self::transWhere($paramName, $paramValue, $old_locale)->first();
            if ($news) {
                $news->translateContext($new_locale);
                $newParams[$paramName] = $news->$paramName;
            }
        }
        return $newParams;
    }


LukeTowers:
I'm not sure that this is the best solution, as I assume that this plugin already handles passing the correct slug based on the language setting, it's just that the localePicker fails to switch to the correct slug when changing language? Is that assumption correct?

mariavilaro:
yes that is the exact problem, the localePicker translates the url but not the slug

mariavilaro:
I was thinking about this, and I can only think of this way to do it. How else can you ask a Model for the translated url parameters? It's now always a slug, sometimes it's a category and a slug, and sometimes a wildcard parameter. For example, I have a model for a search engine with SEO urls like this:

    url = "/search/:whatever*"

And this "whatever" can be:

city
city/family
city/operation
city/operation/family

I have to search in the db to know if the second param is one or the other, and for example the city part is not translated but the family and operation are.

So with my solution, we make datepicker throw an event to let the plugin know that we are translating the url so the plugin can give us the translation.

LukeTowers:
Ah, so the Translate plugin already requires plugins to handle the URL param translation themselves?

In that case, then yes, the event makes the most sense. We'll have to get @daftspunk's opinion on the event first though, once an event has been added it's next to impossible to remove or change it because code in the wild will be using the event; so it's important to get it as right as possible before adding new events

LukeTowers:
Related: #167

LukeTowers:
@mariavilaro could you please rebase this pull request against master and only include your event to translate page params? After reading the other issues I'm liking this approach the most and I'd like to merge it.

mariavilaro:
It's ok now?

LukeTowers:
Looks good, could you also make an edit to the README documentation to document how to use this event?

CptMeatball:
So basically you need to listen to an event in your model, and manually translate that in the model? Seems a bit like a patch, rather than a actual fix tbh.

Everything is handled by the translatable behaviour, except the translation of a field when used as a parameter?

Might be looking at this wrong, but AFAIK you use the slugs that match the locale. Having one slug in one locale, and another in a different locale looks like an edge case to me. 

Anyway, not sure if possible but looks like it would make more sense if this is implemented in the behaviour rather as chunk of code that needs be added to every multilingual model with a slug. 

Or am I really failing to see the logic in this approach?

LukeTowers:
@CptMeatball the translatable model doesn't know which attributes of a model are used within a URL or even when a particular URL has anything to do with a model at all. Typically, URL parameters are used by Components to load specific data. As such, it should be the responsibility of the component (plugin) author to ensure that their components can support multilingual URL parameters if they have added the backend support to define those parameters in multiple languages.

Thoughts?

LukeTowers:
After a lengthly discussion with @CptMeatball in Slack (#development), here's the ideal implementation that we've come up with. @mariavilaro if you could give me some feedback on the idea, that would be much appreciated!

Basically the problem boils down to the fact that URL parameters need to be able to be translated dynamically when switching locales. URL parameters are used by components, so I'm thinking we make this as simple as possible to developers by allowing them to only have to specify some additional configuration options when defining their component properties.

Example: (using [RainLab.Blog's Post component](https://github.com/rainlab/blog-plugin/blob/master/components/Post.php#L28))
```php
public function defineProperties()
{
    return [
        'slug' => [
            'title'       => 'rainlab.blog::lang.settings.post_slug',
            'description' => 'rainlab.blog::lang.settings.post_slug_description',
            'default'     => '{{ :slug }}',
            'type'        => 'string',
            'translatable' => [
                'model' => \RainLab\Blog\Models\Post::class,
                'property' => 'slug',
            ],
        ], ...
    ];
}
```

The LocalePicker would now scan a page's component's properties to find the URL parameters that are translatable (along with the information required to actually translate them):
```php
...

$components = $page->getComponents() // pseudo code, probably not how you actually get the components on a page / layout

$translatableParams = [];
foreach ($components as $component) {
    $properties = $component->defineProperties();
    foreach ($properties as $propertyKey => $property)
        if (!empty($property['translatable']) {
           $paramName = $component->getParamName($propertyKey); // https://github.com/octobercms/october/blob/develop/modules/cms/classes/ComponentBase.php#L191
           if ($paramName) {
               $translatableParams[$paramName] = $property['translatable'];
           }
        }
    }
}
```
This would build an array containing component properties that are parameters for the current page along with the information required to translate those properties:
```php
$translatableParams = [
    'slug' => [
        'model' => \RainLab\Blog\Models\Post::class,
        'property' => 'slug',
    ], ...
]
```

This could then be used by the router to translate all the URL params:
```php
$translatedParams = [];
foreach ($translatableParams as $param => $config) {
    $translatedParams[$param] = $config['model']::translateProperties([$config['property']], $old_locale, $new_locale)[$config['property']];
}

$params = array_merge($params, $translatedParams);
```

The TranslatableModel behavior would include a new static helper method:
```php
public static function translateProperties($properties, $sourceLocale, $targetLocale)
{
    foreach ($properties as $property => $value) {
        $model = self::transWhere($property, $value, $sourceLocale);
        if ($model) {
            $model->translateContext($targetLocale);
            $properties[$property] = $model->$property;
        }
    }
    return $properties;
}
```

mariavilaro:
Looks good, but I tested it with Blog Plugin and I found a problem that I don't know how to solve. This code works in LocalePicker makeLocaleUrlFromPage method:
```php
$components = $page->components;

$translatableParams = [];
foreach ($components as $component) {
    $properties = $component->defineProperties();
    foreach ($properties as $propertyKey => $property) {
        if (!empty($property['translatable'])) {
           $paramName = $component->paramName($propertyKey);
           if ($paramName) {
               $translatableParams[$paramName] = $property['translatable'];
           }
        }
    }
}

$translatedParams = [];
foreach ($translatableParams as $param => $config) {
    $translatedParams[$param] = $config['model']::translateProperty($config['property'], $params[$config['property']], $this->oldLocale, $locale);
}

$params = array_merge($params, $translatedParams);
```

But this doesn't work in TranslatableModel behaviour, while it works as a method in Post model:

```php
public static function translateProperty($property, $value, $sourceLocale, $targetLocale)
{
    $model = self::transWhere($property, $value, $sourceLocale)->first();
    if ($model) {
        $model->translateContext($targetLocale);
        $value = $model->$property;
    }

    return $value;
}
```

Error: "Call to undefined method RainLab\Translate\Behaviors\TranslatableModel::transWhere()" on line 198 of C:\wamp\www\utopigstudio\plugins\rainlab\translate\behaviors\TranslatableModel.php

LukeTowers:
Try `static::transWhere()` instead

munxar:
For me the relation **property <-> model** seems to be too coupled.
If I want to translate a property, only the component should be involved in the interface.
Reason: I have several cases where multiple models are involved in the translation e.g. a search result where each type can have a different model or no model at all.

```php
public function defineProperties()
{
    return [
        'slug' => [
            'title'       => 'rainlab.blog::lang.settings.post_slug',
            'description' => 'rainlab.blog::lang.settings.post_slug_description',
            'type'        => 'string',
            'default'     => '{{ :slug }}',
            'translatable' => ['slug'],
        ], ...
    ];
}

// implemented in component, the picker would directly call that
public static function translateProperty($property, $value, $sourceLocale, $targetLocale)
{
    // here I can decide what model(s) are involved (if at all, no model could be possible)
    $model = MyModel::transWhere($property, $value, $sourceLocale)->first();
    if ($model) {
        $model->translateContext($targetLocale);
        $value = $model->$property;
    }

    return $value;
}
```
We just tell what url parameters we want to be translatable and then translate them in the component. I know that this is not so convenient, but imho much more flexible.
To give back the convenient part, we could implement a trait that is applied to the component, that does implement translateProperty bound to one model, like suggested.
The component could define the translatable properties like so

```php
class MyComponent extends Component {
    use TranslatableProperties;
    public $translatableProperties = ['slug' => MyModel::class]
}
```

We could even drop the 'translate' definition inside of the defineProperties() method, if we use a little syntactic sugar like that:

```php
class MyComponent extends Component {
    use TranslatableProperties;
    public $translatableProperties = [
        'slug' => MyModel::class,                               // uses class MyModel with attribute slug to translate
        'foo' => [OtherModel::class => 'foo_slug'],   // uses class OtherModel with attribute foo_slug to translate
        'bar'                                                                  // calls the translateProperty method in the component
   ]
   public static function translateProperty($property, $value, $sourceLocale, $targetLocale)
   {
        // implement me
        return 'yada yada';
   }
}
```

LukeTowers:
@munxar good suggestions! I'd agree with you that we shouldn't be coupling the model so closely with the component properties. How can we make the developer experience as smooth as possible with such a decoupling approach though?

I'd like to avoid having to define a separate property for listing the properties that are translatable, we should be able to achieve that within the defineProperties() method:
```php
public function defineProperties()
{
    return [
        'slug' => [
            'title'       => 'rainlab.blog::lang.settings.post_slug',
            'description' => 'rainlab.blog::lang.settings.post_slug_description',
            'type'        => 'string',
            'default'     => '{{ :slug }}',
            'translatable' => true, // calls translateProperty in component
            'translatable' => MyModel::class, //  uses class MyModel with attribute slug to translate
            'translatable' => [OtherModel::class => 'foo_slug'], // uses class OtherModel with attribute foo_slug to translate
        ], ...
    ];
}
```
What do you imagine the implementations using the models to translate the properties looking like? Were you thinking that use case would use the above translateProperty() method on the model while the `true` use case would use the translateProperty() method on the component?

munxar:
> I'd like to avoid having to define a separate property for listing the properties that are translatable
The point is, consistency. The TranslatableModel works like that and translation is something you could optional add or remove.
If we add that to the normal property config, the chances of breaking existing code is higher.
let's try to stick to the translatable model:

```php
class MyModel extends Model {
    public $implements = [TranslatableModel::class];
    public $translatable = ['title', 'slug' => ['index' => true]];
}
```
I try to transfer this as close as possible to components:

```php
class MyComponent extends Component {
    public $implements = [TranslatableComponent::class];
    public $translatable = ['slug', 'foo' => ['model' => MyModel::class], 'bar' => ['model' => OtherModel::class, 'name' => 'baz']];
}
```
this is all you add to new or existing models. we don't have to change the documentation or handling of the base component class.

CptMeatball:
Not sure I agree with the initial reason / problem the proposed approach wouldn't work.

I hadn't though that in some cases a slug can belong to different models. Which is, in hindsight, kind of feasible. But I think that's non-standard behaviour and also the job of the component to sort out which model has to be used in what occasion. 

In that specific scenario it would seem more practical to let a function decide which model to use:
```
[..]
'slug' => [
  'translatable' => [
    'model' => $this->getSlugModel();
  ]
]
[...]
public function getSlugModel(){
  // return Model::class
}
```

Also, I think we shouldn't forget that this is only for the `localePicker` language switch. There are already `transWhere` methods in place that work in these scenarios. Also, the functionality for multiple models : 1 slug is already in place inside of the component and has nothing to do with the model.

munxar:
> .. in which he said that's not the right place to do so
please explain why, I brought my arguments
1. consistency with existing interface
2. don't touch existing code

LukeTowers:
@munxar I'm less concerned about consistency with the existing interface in this case because it's for a completely different section. What exactly does the `TranslatableComponent` behaviour do in your example?

As for the `don't touch existing code/properties`, adding extra options to the property definitions is perfectly fine, October is built to handle that if only for the reason to potentially support different property types and their configurations in the future. You'd have to provide an example of real code that would break from an additional element being added to those property definitions.

My reasoning for using the existing defined properties and not a separate class property is because the URL parameters can only be used in a component property. There is no case where someone needs to translate a component property that isn't defined in defineProperties, so it doesn't really make sense to force the developer to have to define a property twice in the same file.

Thoughts?

LukeTowers:
@mariavilaro @CptMeatball @munxar Any further thoughts on this discussion?

CptMeatball:
Hmm, I don't see any problems with your reasoning Luke. I think it's a good approach to use this in the component in the way you described. What's Samuell's thoughts on this?

LukeTowers:
@CptMeatball he's a bit busy right now with the upcoming push to stable of Build 420+ but I'll ask him while we're coordinating that

munxar:
@LukeTowers 
I would say, we should merge this PR and discuss any more extensions in a different PR.
It's the base for whatever we implement next, and to be honest, I almost wrote the same code to make slugs translate on locale switch.

LukeTowers:
@munxar I'd need @mariavilaro to add documentation for the event inline and in the readme before I can merge this.

As another note, I thought of a reason why we'd want to have a different property for the translatable properties, because then we could extend a plugin from another plugin to add support for translated properties without requiring the original author to properly support it.

datune:
@LukeTowers @mariavilaro What's the current status for this issue? Anything I can do to get this merged?

mjauvin:
#341 makes this obsolete.

PR #266, opened 2017-05-31
storeTranslatableIndexData returning within a loop

thehiggster:
When updating a model, the storeTranslatableIndexData is called. Inside the foreach loop, if any of the attributes are empty strings, on line 151, it returns and drops out. If the next attribute is not empty, and needs to be updated, it fails. Changing this to a continue resolves this. 

Don't believe I'm missing anything but if there's a better solution or a reason for the return, happy to change!
Thanks, 
James

### DIFF
--- a/behaviors/TranslatableModel.php
+++ b/behaviors/TranslatableModel.php
@@ -128,7 +128,6 @@ class TranslatableModel extends TranslatableBehavior
         }
 
         $data = $this->translatableAttributes[$locale];
-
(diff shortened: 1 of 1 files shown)


munxar:
I confirm that this fixes the described issue for me.

LukeTowers:
I'll merge this, but I haven't confirmed the original reported bug in the first place personally. Thanks for your input @munxar! @daftspunk let me know if that `return` was supposed to be a `return` and not a `continue`.

PR #272, opened 2017-06-21
search feature

dev-lav:
Added search feature to make it easier for users with lots of data to translate.
Thank you.

### DIFF
--- a/controllers/Messages.php
+++ b/controllers/Messages.php
@@ -71,6 +71,7 @@ class Messages extends Controller
 
     public function prepareTable()
     {
+        $search   = input('search', null);
         $fromCode = post('locale_from', null);
         $toCode = post('locale_to', Locale::getDefault()->code);
         $this->hideTranslated = post('hide_translated', false);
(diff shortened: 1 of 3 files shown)


LukeTowers:
I don't have a ton of time to review this right now, so if somebody was able to take screenshots of this in action (or even better a GIF) as well as confirm that it works, I'd love to merge it.

Thanks for your work so far @dev-lav!

daftspunk:
As above. Thanks @dev-lav

jan-vince:
Hi, I have tried this, but have no time to dig much deeper:

* 'Search ...' should be replaced with translation ```` e(trans('backend::lang.list.search'))````.
* Search should use AJAX to be consistent across October backend (right now I have to hit Enter key to search).
* For me, search works only for default language. 
* After search, list is reversed to default view (so if I choose different language in list header, after search it's back in default).

What do you think about these points?

LukeTowers:
@dev-lav See if you can integrate the [search widget](https://github.com/octobercms/october/blob/master/modules/backend/widgets/Search.php) in a similar fashion to how the ListsController behaviour does already. If the docs are sparse on it, try to figure it out as much as possible, and then I would very much welcome a PR to add more inline documentation to the code so that the [API documentation for that widget](http://octobercms.com/docs/api/backend/widgets/search) would be improved.

daftspunk:
Added seaching to the table widget instead fb007dfb7b7968df8f766c7da5b4c0d565a81841

Note Translate 1.3.0+ requires Build 420+

daftspunk:
BTW - great contribution @dev-lav!

dev-lav:
Thank you @daftspunk

PR #287, opened 2017-08-10
Sortable Language List

munxar:
Due to the repeating requirement that languages should be in a customer defined order, this PR adds exactly this.
The sort order is reflected in the form widgets and the localePicker.


LukeTowers:
Why'd you close this?

munxar:
@LukeTowers I can resolve a problem that will happen if someone installs this plugin on a new instance. Let me explain what the problem is, eventually you know how to solve this:
The first entry in the version.yaml sets up all tables an seeds the default language en.
Now, if the seed runs, it uses my version of the Locale model with the newly added Sortable trait.
When the seed file creates the default locale, the added trait kicks in and tries to add value to the  sort_order column. But this column doesn't exist at this point in time.
Imho this is a conceptual problem with seeds.
Note: I accidentally deleted the repo for this pull request but made a new one here:
https://github.com/rainlab/translate-plugin/pull/289

munxar:
while I wrote you, I found a potential solution -> #289

PR #289, opened 2017-08-10
Added reorder to languages

munxar:


munxar:
Due to the repeating requirement that languages should be in a customer defined order, this PR adds exactly this.
The sort order is reflected in the form widgets and the localePicker.
Note: right now it doesn't work, if installed from scratch.
Problem: Sortable trait is uses when migrating the plugin from version 1.0.1 up, but the column sort_order dosn't (yet) exist.

LukeTowers:
@munxar if another person could independently test and verify this I'll merge it as I don't have time to test it myself. Looks good though!

tschortsch:
Tested it on a project where I'm using the translate plugin and it worked well. 👍

LukeTowers:
Awesome, thanks @tschortsch. Great job @munxar!

PR #284, opened 2017-08-02
Fix problems with the unavailability of the sessions in the App::before

hocopor:
Fixes #283, Fixes #282

PR #291, opened 2017-08-15
Add missing web middleware

acasar:
When upgrading to Laravel 5.5 I got an error when visiting the homepage in a foreign locale: "Session store not set on request." Adding the "web" middleware to one remaining route fixed it.

I believe this was just an oversight in this commit: https://github.com/rainlab/translate-plugin/commit/aa6e523dbe47c60e6ee318e3a7e1a32d7ca52db8

### DIFF
--- a/routes.php
+++ b/routes.php
@@ -34,7 +34,7 @@ App::before(function($request) {
         Route::any('{slug}', 'Cms\Classes\CmsController@run')->where('slug', '(.*)?');
     });
 
-    Route::any($locale, 'Cms\Classes\CmsController@run');
+    Route::any($locale, 'Cms\Classes\CmsController@run')->middleware('web');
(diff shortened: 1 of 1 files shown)


LukeTowers:
Hmm, does this solve the problem that #284 is trying to solve? One of these two solutions needs to be decided upon.

acasar:
@LukeTowers I don't think this is related to https://github.com/rainlab/translate-plugin/pull/284

This PR only makes sure that session is available on all routes, which we need to regardless of https://github.com/rainlab/translate-plugin/pull/284.

LukeTowers:
Well I'll merge it for now and we can decide later whether to go with #284 or if this fixes the remaining issues with Laravel 5.5

PR #299, opened 2017-09-21
Support maxlength attribute

mariavilaro:
I moved this change to its own branch, please close #199 

This is a small change that allows to define a maxlength attribute for a multilanguage text field


LukeTowers:
Closing because #300 includes these changes.

mariavilaro:
Yeah I moved it here in it's own branch because it's a very small change that maybe had more chances to go live

Issue #87, opened 2015-08-20
Can't translate Mail Templates


zek:
I've tried to translate mail templates but I couldn't.

Is this plugin supports translating mail templates?

gergo85:
I can't translate the mail templates too so I use different template for every language. See the following plugin for details: https://github.com/gergo85/oc-news

wellyfrs:
Same problem.

LukeTowers:
If anyone wants to work on implementing a fix for this, please do so and make a PR. Otherwise, just adding +1 or "same problem" is not helpful.

wellyfrs:
All right, @LukeTowers. I made one request (#304), but it depends on another one in October (https://github.com/octobercms/october/pull/3152). With this, the search also scans e-mail templates.

LukeTowers:
@wellyfrs awesome thanks, will wait to merge #304 until after merging octobercms/october#3152

PR #304, opened 2017-10-04
Adding method to scan translatable messages into email templates

wellyfrs:
Adding method to scan translatable messages into email templates. See: https://github.com/rainlab/translate-plugin/issues/87. It depends: https://github.com/octobercms/october/pull/3152


LukeTowers:
Will wait for octobercms/october#3152 before merging.

Issue #315, opened 2017-10-31
Add setting to disable retrieving locale from session


vannut:
For almost all of my projects I want a slightly different use of the routing of the Translate plugin. I found a solution, but before making a PR I want to discuss it here with you.

### Current behaviour:
At the moment when you visit an url without a locale-prefix in the url-segments the (new) [localeMiddleware](https://github.com/rainlab/translate-plugin/blob/master/classes/LocaleMiddleware.php#L20) looks for the locale stored in the session. 

So domain.tld/ will always show the default page in the **last** visited locale. So if i first visited `domain.tld/de` and afterwards I would go to `domain.tld/` the page would be in German(de). 

### Desired behaviour
When no locale found in the url::segment(1), just stick to the default locale. Result would be the following:
- domain.tld/  <-- always the default locale
- domain.tld/es  <-- alternate locale
- domain.tld/nl <-- alternate locale 2
- domain.tld/en <-- default locale

### Solution:
Make the retrieval of the locale from the session optional. 
```php
// Current LocaleMiddleware.php L:20
if (!$translator->loadLocaleFromRequest()) {
    $translator->loadLocaleFromSession();
}
```
Commenting out this if -statement results in the desired behaviour. So I would like to add a setting somewhere which disables this statement. 

As we not yet have a settings-model for this plugin I would discuss if we add one. But that might be a bit overkill for just one setting. 
We could add a config setting in /config/translate.php which defaults to true (when not found) so we have the possibility to add it when needed....

vannut:
@datune This adresses your 'problem' with my plugin.

datune:
@vannut Thanks! 

In fact, your described "desired behaviour" is the best practice for multilingual sites. Google and co. also strongly discourages setting site language based on cookies without an explicit domain or uri parameter. 

The default behaviour of the Rainlab.Translate plugin is actually a bad practice, especially for SEO. 

For those who want to know more: [read this](https://support.google.com/webmasters/answer/182192)

CptMeatball:
I feel like skipping the session entirely is somewhat of an edge case. 

That said, I think it _should_ be possible, but not through backend. I'd favour a settings file over a new settings tab with only two checkboxes. Something like:

Settings/translate.php
```
/*
*  null => current behavior
*  true => Skip Session / Force URL
*/
`skipSession` => `null`,
`forceLocale` => 'null'
```

==

[...]

PR #317, opened 2017-11-02
Adding the possibility to skip session retrieval

vannut:
As discussed in #315 this is a solution to add a setting which skips the retrieval of the locale from the session. 
By the default the behaviour of the plugin is not changed; when visiting the `domain.tld/` route it wil display the __last loaded__ locale. 
Setting `skipSession = true` will disable this and keep the locale on the default locale. Result is a SEO friendly domain-structure; where every locale has one, and only one unique url-path.  

To activate this add an entry in `/config/cms.php`
```php
...
'translate' => [
    'skipSession' => true
]
...
```

### DIFF
--- a/classes/LocaleMiddleware.php
+++ b/classes/LocaleMiddleware.php
@@ -17,7 +17,8 @@ class LocaleMiddleware
         $translator = Translator::instance();
         $translator->isConfigured();
 
-        if (!$translator->loadLocaleFromRequest()) {
+
+        if (!$translator->loadLocaleFromRequest() && !$translator->skipSessionLocaleRetrieval()) {
             $translator->loadLocaleFromSession();
         }
 
--- a/classes/Translator.php
+++ b/classes/Translator.php
@@ -91,7 +91,7 @@ class Translator
     }
 
     /**
-     * Check if this plugin is installed and the database is available, 
+     * Check if this plugin is installed and the database is available,
      * stores the result in the session for efficiency.
      * @return boolean
      */
@@ -208,4 +208,17 @@ class Translator
     {
         Session::put(self::SESSION_LOCALE, $locale);
     }
+
+    /**
+     * Should we load the locale from the session?
+     * See https://github.com/rainlab/translate-plugin/issues/315
+     *
+     * @return boolean
+     */
+    public function skipSessionLocaleRetrieval()
+    {
+        return config('cms.translate.skipSession', false);
+
+    }
+
 }
--- a/classes/Translator.php
+++ b/classes/Translator.php
@@ -2,6 +2,7 @@
 
 use App;
 use Schema;
+use Config;
 use Session;
 use Request;
 use RainLab\Translate\Models\Locale;
@@ -217,8 +218,7 @@ class Translator
      */
     public function skipSessionLocaleRetrieval()
     {
-        return config('cms.translate.skipSession', false);
-
+        return Config::get('rainlab.translate::skipSession', false);
     }
 
 }
--- a/config/config.php
+++ b/config/config.php
@@ -1,0 +1,17 @@
+<?php
+
+return [
+
+    /*
+    |--------------------------------------------------------------------------
+    | Skip Session
+    |--------------------------------------------------------------------------
+    |
+    | When using the rainlab.Translate plugin it sometimes is necessary to
+    | skip te retrieval of the locale from the session in order to let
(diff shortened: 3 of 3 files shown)


LukeTowers:
@vannut config/cms.php is for October, not for plugins. Please create a plugin specific config file that would then be overridden by config/rainlab/translate/config.php. See http://octobercms.com/docs/plugin/settings#file-configuration

vannut:
Ow. Still discovering new things within October :)

vannut:
I didn't/don't use the localeSwitcher provided by the plugin. As a result I do not use the checkbox to force 'the URL-schema. 
To change the language of the website i change the  url to include the locale as a prefix. I made a [component](https://github.com/vannut/oc-trans-ext-plugin/tree/master/components) which changes the url directly based on the current page. Which doesn't rely on the session anymore :)

Bullet 2 is expected.

datune:
@vannut That's why I tested your PR without your plugin installed, since the majority of current users will use the localeSwitcher from Rainlab.Translate. 

Based on your PR description I don't see how bullet 2 is expected? I might be sleepy or missing the obvious ;-)

I expected for this PR to ensure that, if the URL Schema is enforced, it will only do so for all languages other than the default one?

vannut:
> I expected for this PR to ensure that, if the URL Schema is enforced, it will only do so for all languages other than the default one?

Well actually it does something else, with eventually the same outcome.

By default _every_ locale has its own locale-prefixed domain structure: `domain.tld/nl/...` `domain.tld/en/...` and `domain.tld/de/...` regardles of what the default-locale is. They exists for all the enabled-locales.

But Rainlab.Translate also has the non-prefixed url-structure: `domain.tld/...`

There are two settings for what should happen with this non-prefixed domain-structure: 
1 - Without the force schema setting; it simply displays the page in the last locale known in the session. (bad for seo as the root can have multiple locales)
2 - With the force schema setting enabled; it wil redirect to the last known locale in the session, with a prefix for every locale. (Bad as it results in `domain.nl/nl/..` for default dutch websites.)

What the setting in this PR does  --when you *don't* force the schema-- is to make sure that the non-prefixed url is still the default locale and not the session locale. 

Does that make any sense. 

Regardless; we indeed need a solution for the localePicker.

datune:
Alright, so basically, those who are currently using non prefixed language switching (which relies on the sesssion) should never set skipSession to true. This ensures bc, which, as implemented right now should work. 

For those who want SEO friendly multi language sites, we need a solution. Since it's obvious that this requires force URL schema to be enabled, it seems to me that we need to address the fact that it enforces the url schema for all languages, when really, it should only enforce it for all languages except the default one. 

To elaborate: A website, foobar.de with two languages, english (en) and german (de), having german set to default with force URL Schema and skipSession set to true

**Current behaviour**
foobar.de/de will display the german language
foobar.de will also display the german language but will never be shown, since it always redirects to foobar.de/de
foobar.de/en will display the english language

**Expected behaviour:** 
foobar.de -> will display the default language (de)
foobar.de/de will redirect to foobar.de, since it is the default language
foobar.de/en -> will display english

Your current PR is the first step, ensuring that the non-prefixed domain always loads the default language.

vannut:
> For those who want SEO friendly multi language sites, we need a solution. 

What am I missing? If you don't set `force_url` and `skipSession=true` the domain tree is correct.
For seo purposes we just need to add a `hreflang element` with all the alternate locales. 
I'll throw in an other PR with that component.

Switching to an other locale by means of the localePicker apparently depends on the session stored locale.

I'm a little bit lost in what to do next.

datune:
@vannut 

If you don't set force_url, you'll never get domain prefixes.

vannut:
That's not true. 
Regardless of the setting in the localepicker-property there will be a tree per locale with the prefix!

datune:
> That's not true.
Regardless of the setting in the localepicker-property there will be a tree per locale with the prefix!

I think you misunderstand. If you don't set force_url, you'll never get redirects to domain prefixes. Which is the purpose of force url to begin with.

vannut:
That's correct!! 
Lets discuss in Slack what you really mean.  :)

vannut:
All right. Back to the drawing board. 
Although the proposed setting works for always displaying the default locale on the /-root; getting it to work with the redirect magic of `force url schema` (and the locale-picker itself), without causing a bc is more complex. 

I'm closing this PR for now and will be back with a new one soon.

PR #318, opened 2017-11-06
PrefixDefaultLocale setting and Alt hreflang element

vannut:
This PR adds two things to the translate plugin:
1 - A component which renders alt hreflang elements 
2 - adds a setting to ask if we need to prefix the default locale

## Alternate hreflang elements
According to [Google's locale recommendations](https://support.google.com/webmasters/answer/189077) we need to inject alternative languages in the head of a website as following:
```html
<link rel="alternate" hreflang="en" href="/en" />
<link rel="alternate" hreflang="nl" href="/nl" />
```
This needs to be done for _every_ language including the current language.

When creating this component, the fact I didn't want to have the default locale prefixed, issue #315 regarding the behaviour when redirecting to the default locale and PR #317 let me to alter the behaviour of the forceurl redirect-method, the second part of this PR

## prefixDefaultLocale
You can override the setting locally in [your own config file.](http://octobercms.com/docs/plugin/settings#file-configuration)
What it does: 
When having `forceURL schema = true` and `prefixDefaultLocale=false` the website wont redirect the default locale to `/default_locale/...` Instead it just remains on the main-root.
To make sure the main root does not display the session-stored locale; the session locale gets reset in the same method :)
If you leave `prefixDefaultLocale=true` nothing should change :)

### DIFF
--- a/classes/Translator.php
+++ b/classes/Translator.php
@@ -91,7 +91,7 @@ class Translator
     }
 
     /**
-     * Check if this plugin is installed and the database is available, 
+     * Check if this plugin is installed and the database is available,
      * stores the result in the session for efficiency.
      * @return boolean
      */
@@ -151,9 +151,10 @@ class Translator
      *
      * @param string $path Path to rewrite
      * @param string $locale optional language code, default to the system default language
+     * @param boolean $removeDefaultLocale dont prefix the path with the default locale
      * @return string
      */
-    public function getPathInLocale($path, $locale = null)
+    public function getPathInLocale($path, $locale = null, $removeDefaultLocale = false)
     {
         $segments = explode('/', $path);
 
@@ -171,6 +172,12 @@ class Translator
         else {
             array_unshift($segments, $locale);
         }
+
+        if ($removeDefaultLocale
+            && isset($segments[0]) && $segments[0] == $this->defaultLocale
+        ) {
+            array_shift($segments);
+        };
 
         return implode('/', $segments);
     }
--- a/classes/Translator.php
+++ b/classes/Translator.php
@@ -184,7 +184,7 @@ class Translator
             && isset($segments[0])
             && $segments[0] == $this->defaultLocale
         ) {
-            // remove the first element (whichis the default locale)
+            // remove the first element (which is the default locale)
             array_shift($segments);
         };
 
--- a/classes/Translator.php
+++ b/classes/Translator.php
@@ -4,6 +4,7 @@ use App;
 use Schema;
 use Session;
 use Request;
+use Config;
 use RainLab\Translate\Models\Locale;
 
 /**
@@ -149,13 +150,17 @@ class Translator
     /**
      * Returns the path prefixed with language code.
      *
-     * @param string $path Path to rewrite
+     * @param string $path Path to rewrite, already translate, with or without locale prefixed
      * @param string $locale optional language code, default to the system default language
-     * @param boolean $removeDefaultLocale dont prefix the path with the default locale
+     * @param boolean $prefixDefaultLocale should we prefix the path when the locale = default locale
      * @return string
      */
-    public function getPathInLocale($path, $locale = null, $removeDefaultLocale = false)
+    public function getPathInLocale($path, $locale = null, $prefixDefaultLocale = null)
     {
+        $prefixDefaultLocale = (is_null($prefixDefaultLocale))
+            ? Config::get('rainlab.translate::prefixDefaultLocale')
+            : $prefixDefaultLocale;
+
         $segments = explode('/', $path);
 
         $segments = array_values(array_filter($segments, function ($v) {
@@ -168,14 +173,18 @@ class Translator
 
         if (count($segments) == 0 || Locale::isValid($segments[0])) {
             $segments[0] = $locale;
-        }
-        else {
+        } else {
             array_unshift($segments, $locale);
         }
 
-        if ($removeDefaultLocale
-            && isset($segments[0]) && $segments[0] == $this->defaultLocale
+        // If we don't want te default locale to be prefixed
+        // AND the first segment equals the defaultlocale
+        if (
+            !$prefixDefaultLocale
+            && isset($segments[0])
+            && $segments[0] == $this->defaultLocale
         ) {
+            // remove the first element (whichis the default locale)
             array_shift($segments);
         };
 
--- a/components/AlternateHrefLangElements.php
+++ b/components/AlternateHrefLangElements.php
@@ -36,7 +36,7 @@ class AlternateHrefLangElements extends ComponentBase
     {
         $translator = Translator::instance();
         $page = $this->getPage();
-        
+
         /*
          * Static Page
          */
@@ -55,7 +55,7 @@ class AlternateHrefLangElements extends ComponentBase
             $localeUrl = $router->urlFromPattern($page->url, $params);
         }
 
-        return $translator->getPathInLocale($localeUrl, $locale, true);
+        return $translator->getPathInLocale($localeUrl, $locale);
     }
 
 }
--- a/components/AlternateHrefLangElements.php
+++ b/components/AlternateHrefLangElements.php
@@ -1,0 +1,61 @@
+<?php
+
+namespace Rainlab\Translate\Components;
+
+use Cms\Classes\ComponentBase;
+use RainLab\Translate\Classes\Translator;
+use RainLab\Translate\Models\Locale as LocaleModel;
+use October\Rain\Router\Router as RainRouter;
+
+class AlternateHrefLangElements extends ComponentBase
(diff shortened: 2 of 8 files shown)


classes/Translator.php
```diff
@@ -167,11 +173,21 @@ public function getPathInLocale($path, $locale = null)
...
         }
 
+        // If we don't want te default locale to be prefixed
+        // AND the first segment equals the defaultlocale
+        if (
+            !$prefixDefaultLocale
+            && isset($segments[0])
+            && $segments[0] == $this->defaultLocale
+        ) {
+            // remove the first element (whichis the default locale)
```
LukeTowers: `whichis` to `which is` please.

components/AlternateHrefLangElements.php
```diff
@@ -0,0 +1,61 @@
...
+use October\Rain\Router\Router as RainRouter;
+
+class AlternateHrefLangElements extends ComponentBase
+{
+
+
+    public function componentDetails()
+    {
+        return [
+            'name' => 'Alternate HREF Lang elements',
```
LukeTowers: Provide language keys for this component's name and description please, it would be funny if the Translation plugin didn't actually provide a translation for itself :)

components/alternatehreflangelements/default.htm
```diff
@@ -0,0 +1,3 @@
+{% for loc, url in __SELF__.locales %}
+    <link rel="alternate" hreflang="{{ loc }}" href="/{{ url }}" />
```
LukeTowers: Does this support October living in a subdirectory? Could we process `url` before passing it to the view to get the fully qualified URL instead here?

components/AlternateHrefLangElements.php
vannut: That would be strange indeed ;)
Adding them as we speak!

components/alternatehreflangelements/default.htm
vannut: Never thought about the 'subdirectory' thing. Never used it. 
We could process `url` indeed and prefix it with `env('APP_URL')` or is there a better way to do this? Or does OctoberCMS provide a helper for this?

vannut: Hmm forgot, by default octobercms does not use .env files. So probably prefix it with `Config('app.url')` then instead?

vannut:
Regarding #244 I'll look into that later; have to read (and understand) the issue and thread first. And time is very scarce atm.

components/alternatehreflangelements/default.htm
LukeTowers: Give `url()` a try, let me know how it works

LukeTowers:
@vannut that's fine, we can look at it later, just wanted you to be aware of it. As soon as it supports subdirectories I'll merge this. Great work!

components/alternatehreflangelements/default.htm
```diff
@@ -1,3 +1,3 @@
-{% for loc, url in __SELF__.locales %}
```
LukeTowers: Please revert back to these variables, `l` and `u` are very cryptic and hard to understand. I'd rather have `locale` and `alternateUrl`

LukeTowers: @vannut

vannut:
@LukeTowers I reverted those vars to be more verbose.

components/alternatehreflangelements/default.htm
```diff
@@ -0,0 +1,3 @@
+{% for locale, alternateUrl in __SELF__.locales %}
+    <link rel="alternate" hreflang="{{ locale }}" href="{{ url(alternateUrl) }}" />
```
LukeTowers: What does the `url()` function do? Does it actually work? I've never used it in October. Why not build the full URL in the PHP code that's creating the `alternateUrl` value and just use `{{ alternateUrl }}`?

vannut: It is a helper for  [`URLGenerator->to($path)`](https://github.com/laravel/framework/blob/5.5/src/Illuminate/Routing/UrlGenerator.php#L170)

LukeTowers: So long as it works I'm fine with it then :) Thanks for your work on it! Is this ready to merge now?

vannut: It spits out `https://build426.dev/en/tesssst` instead of `/en/tesssst` :)
So lets merge! If it creates a problem somehow we can adjust it easily.

msimkunas:
The current middleware (`classes/LocaleMiddleware.php`) implementation needs to be updated to respect the existence of the `prefixDefaultLocale` setting like so:

```
public function handle($request, Closure $next)                                                                                                                                  
{                                                                                                                                                                                
    $translator = Translator::instance();
    $translator->isConfigured();
    if( ! $translator->loadLocaleFromRequest()) {
        if(Config::get('translate.prefixDefaultLocale')) {
            $translator->loadLocaleFromSession();
        } else {
            $translator->setLocale($translator->getDefaultLocale());
        }
     }

     return $next($request);
}
```

LukeTowers:
@msimkunas feel free to update the config's comment to better reflect what happens

msimkunas:
Alright. What about the middleware changes I've suggested?

LukeTowers:
I don't use it enough to be able to tell if they're needed or not

msimkunas:
Okay then, lets see if @vannut has anything to say on the matter.

vannut:
You know what the funny thing is. Just yesterday I was working on this :)
Somehow the setting did not propagate through to the ` {{ 'example' | page }}` filter. But after a refresh of the page it did. 
And now I see this issue 👍 with the solution :)

Let me explain the problem:
We've got two locale: `nl` (default) and `de`. 
We've got `forceURL=1` on the localepicker in place and `prefixDefaultLocale=false`

Visiting `domain.com/` displays the following:
> locale = 'nl'
> page-filter: dutch slug, wo/ prefix
> staticpagefilter: dutch url translation, wo/ prefix
> {{'' |_}}-filter: dutch

Then visit `domain.com/de`:
> locale='de'
> page-filter: proper german translation of the slug, with /de/ prefixed.
> staticpagefilter: same as page filter.
> {{'' |_}}-filter: German. 
Nothing strange here.

Now the thing that goes/went wrong:
After visiting `domain.com/de` and going back to `domain.com/` We would have expected everything to return back to the default locale, nl. But this is what happened:
> locale= 'nl'
> page-filter: **german translation _without_ `/de/` prefix**
> staticFilter: proper dutch translation, wo/ prefix
> {{'' |_}}-filter: **German**

So both the translated message filter and the page-filter where behaving strange. Moreover, after a simple refresh of the page, the page was displayed correctly:
> locale = 'nl'
> page-filter: dutch url translation
> staticpagefilter: dutch url translation
> {{'' |_}}-filter: dutch

This indicated to me there was something fishy going on with the session; but I totally forgot to see the middleware :)

I'll create a fix asap.

msimkunas:
@vannut Great, I'm looking forward to it.

acasar:
@vannut How is this supposed to work? Config on the plugin level is not parsed as far as I know and my existing projects broke.

cc @LukeTowers

LukeTowers:
@acasar sounds like something else is going wrong for you @acasar, config on the plugin level should be parsed just fine, I've never had any issues with my SnappyPDF plugin and that exclusively uses config. https://github.com/LukeTowers/oc-snappypdf-plugin

acasar:
@LukeTowers thanks for feedback, I'll check if there is anything working on my side, since this didn't work until I copied the config to the top level.

Still it probably wouldn't hurt to set a default value?

```php
Config::get('rainlab.translate::prefixDefaultLocale', true)
```

acasar:
@LukeTowers Oh, just noticed - the "rainlab.translate::" prefix is missing here:

https://github.com/rainlab/translate-plugin/blob/master/classes/LocaleMiddleware.php#L22

Is this intentional?

LukeTowers:
Good catch @acasar, if you could make a PR to fix that I'll merge it

msimkunas:
@acasar It's unintentional, thanks for bringing it up. I noticed it some time ago but forgot to update this issue.

vannut:
I'm very sorry @acasar for this problem. I tested it with a config file in `/config/translate.php` and changed the value there from true > false. 
Never thought of ditching the file completely and resort to the config file in the plugin. 

There are some other places within the code where I referred to the config setting, wo/ the prefix. I'll submit a PR to change these.

Again, sorry for the caused trouble!

LukeTowers:
@vannut overriding plugin configs should be done in myproject/config/author/plugin/config.php

acasar:
@vannut No problem, just glad we discovered the issue :)

I created a PR myself to speed things along: https://github.com/rainlab/translate-plugin/pull/329

PR #321, opened 2017-11-17
Version bump

vannut:


LukeTowers:
@vannut this won't be a stable update until @daftspunk makes it so, until then you'll need to enable edgeUpdates to download it through the updates screen

Issue #303, opened 2017-10-03
Array & JSON Casting for translatable attributes empty fields


yarbala:
Cast attributes are saved to db and to 'rainlab_translate_attributes' but have empty value for default language in editor

Model.php:
```
public $translatable = ['title', 'fields[test11]', 'fields[test12]', 'fields[test21]'];
protected $casts = [
        'fields' => 'array'
    ];
```

fields.yml:
```
        'fields[test11]':
            label: Test11
            type: text
            tab: 'Content'
        'fields[test12]':
            label: Test12
            type: text
            tab: 'Content'
```

After saving data I have proper values in DB. I open this form for editing and switch to not default language, then back to default - and field becomes empty.

I'm using latest october version. For other field types like Markdown it works same way.

yarbala:
I made quick fix for it, may be you can add something similar to Plugin

file: plugins/rainlab/translate/classes/TranslatableBehavior.php
```
    protected function getAttributeFromData($data, $attribute)
    {
        $keyArray = HtmlHelper::nameToArray($attribute);

        $path = implode('.', $keyArray);
        $val =  array_get($data, $path);

        // For json attributes
        if (is_null($val) && count($keyArray) > 1) {
            $data[$keyArray[0]] =  json_decode($data[$keyArray[0]], true);
            if (!is_null($data[$keyArray[0]])) {
                $val = array_get($data, $path);
            }
        }

        return $val;
    }
```

LukeTowers:
I have no idea what the issue is from your description. Please try to explain it in more detail and more step by step.

yarbala:
@LukeTowers I'm trying to use dynamic forms(different field types for single model depending on model type) in OctoberCMS backend.  I uses attribute casting to store fields in json field https://laravel.com/docs/5.5/eloquent-mutators#attribute-casting. Everything works fine except translations for dynamic fields. getAttributeFromData function didn't works with json attribute casting. This function tries to get array, but field contains json. My fix solves problem - and my app works fine, but I'm not sure - this fix is best solution.

LukeTowers:
This is a complex case. 

In order to aid in our understanding of the problem, please submit [a Pull Request to the test plugin](https://github.com/daftspunk/oc-test-plugin) for investigation that demonstrates the issue. Don't forget to include step by step instructions how how to replicate it _using the test plugin_.

yarbala:
@LukeTowers I crested pull request with steb by step instructions

[...]

PR #322, opened 2017-11-29
Cast attributes are saved to db and to 'rainlab_translate_attributes'…

yarbala:
… but have empty value for default language in editor

Model.php:

public $translatable = ['title', 'fields[test11]', 'fields[test12]', 'fields[test21]'];
protected $casts = [
        'fields' => 'array'
    ];
fields.yml:

        'fields[test11]':
            label: Test11
            type: text
            tab: 'Content'
        'fields[test12]':
            label: Test12
            type: text
            tab: 'Content'
After saving data I have proper values in DB. I open this form for editing and switch to not default language, then back to default - and field becomes empty.

I'm using latest october version. For other field types like Markdown it works same way.

### DIFF
--- a/classes/TranslatableBehavior.php
+++ b/classes/TranslatableBehavior.php
@@ -399,7 +399,18 @@ abstract class TranslatableBehavior extends ExtensionBase
     {
         $keyArray = HtmlHelper::nameToArray($attribute);
 
-        return array_get($data, implode('.', $keyArray));
+        $path = implode('.', $keyArray);
+        $val =  array_get($data, $path);
+
+        // For json attributes
+        if (is_null($val) && count($keyArray) > 1 && isset($data[$keyArray[0]]) && is_string($data[$keyArray[0]])) {
+            $data[$keyArray[0]] =  json_decode($data[$keyArray[0]], true);
+            if (!is_null($data[$keyArray[0]])) {
+                $val = array_get($data, $path);
+            }
+        }
+
+        return $val;
     }
 
     /**


iboved:
Why Pull request has not yet merged? I have the same problem.

LukeTowers:
@yarbala @iboved @amdad 

I haven't merged this in yet because I don't understand what this fix is doing and I don't have time to become acquainted enough with the issue myself to figure out how this fix works. If someone can break it down for me line by line as to how this fix works then I'd be happy to merge it in.

osmanzeki:
Hi everyone,

I've implemented a proof-of-concept with some explanations in the following repo for the fix in this PR:
https://github.com/osmanzeki/octobercms-jsonable-fix

You should be able to run this pretty easily by reading my `README.MD` file in there and by importing my database dump. I will try to capture a few videos or gifs showing the bugs too to make it even clearer in the future.

Cheers,
Oz

@LukeTowers @yarbala @iboved @amdad

osmanzeki:
I noticed by testing the stuff I did in the themes that this approach seems to break a few things. When a field is in a `jsonable` array but is not in the `translatable`, it seems like you can no longer get the value in the "translated" pages.

See the `meta[image]` value in my examples here:
https://github.com/osmanzeki/octobercms-jsonable-fix/blob/master/themes/demo/pages/home.htm#L51

osmanzeki:
I am starting to think that we should not be doing the `json_decode` in the `getAttributeFromData` method as this PR and my test-repo does but perhaps we should be doing the `json_decode` earlier and giving the data already decoded to the method instead.

Let me know if you guys have some input on this.

mjauvin:
What are "cast" attributes? Never heard of that. @osmanzeki did you make any progress fixing the bug you referenced in your last post?

mjauvin:

yarbala:
@mjauvin https://laravel.com/docs/5.0/eloquent#attribute-casting same to „jsonable” from october

mjauvin:
@osmanzeki first of all, thanks for the very nice and complete description of the issue.

Would it be possible to do the json decoding here:

https://github.com/osmanzeki/octobercms-jsonable-fix/blob/master/plugins/rainlab/translate/classes/TranslatableBehavior.php#L160

With the proper conditions added (like check if the part without the index is in $jsonable or $casts property? i.e. meta[title] -> meta... is meta  in $jsonable or $casts property, do the decoding)

mjauvin:
Something like this maybe (not tested):

```
diff --git a/classes/TranslatableBehavior.php b/classes/TranslatableBehavior.php
index 6cfc3fa..848b221 100644
--- a/classes/TranslatableBehavior.php
+++ b/classes/TranslatableBehavior.php
@@ -160,7 +160,7 @@ abstract class TranslatableBehavior extends ExtensionBase
         if (
             is_string($result) &&
             method_exists($this->model, 'isJsonable') &&
-            $this->model->isJsonable($key)
+            ($this->model->isJsonable($key) || $this->model->isJsonable(HtmlHelper::nameToArray($key)[0]))
         ) {
             $result = json_decode($result, true);
         }
```

osmanzeki:
Hey Marc,

Sorry for the late reply, I have been out of the country for work for the last month or so, just came back to my senses here. :)

I would need to immerse myself back into the problem to be sure for your question. I've added a reminder in my calendar to try to do that this week or the next one.

That said, I think the problem here was that there was a few missing links between all the systems that the translation module interacts with and thus that trying to do the json decoding too late down the function execution order was leading to problems. I'm not sure if this location is early enough to alleviate the issues down the line.

Especially when something is translatable but not a string/text like the image field, file field, another json like the Repeaters, etc.

I think with more test coverage on the rainlab translate plugin we could be better suited to identify where the chain seems to break sometimes.

mjauvin:

mjauvin:
fix for #303

mjauvin:
Closed in favor of simpler fix in #556

Issue #323, opened 2017-12-12
Switching locale of ML repeater doesn't load values of the target locale


Fl0Cri:
I have a page (from Rainlab Static Pages plugin) using a layout that defines a repeater.
I have set some values in the repeater for both languages I am using: EN (default) and FR.
The problem occurs when I try to switch the repeater from the default locale to any other locale.

#### Expected behaviour
When switching to an other locale (here from EN to FR), the fields inside the repeater should be filled with the values from the FR page:
![ml_repeater-working](https://user-images.githubusercontent.com/16371551/33882250-c74b7a86-df37-11e7-93a6-7bc3fd66eab7.png)

#### Actual behaviour
When switching to an other locale, the locale is switched, but the values inside the repeater are the one from the same as the previous locale:
![ml_repeater-not-working](https://user-images.githubusercontent.com/16371551/33882748-7871e7b8-df39-11e7-8150-34448f40a532.png)
But actually, if I look directly inside the static page file, the FR page has the correct values (`Repeater FR ...`, etc.).
If I hit save the save button in the backend when in this situation, all the values of the repeater in the FR page are overwritten by the EN values (what is displyed on the screen).

#### Reproduce steps
1. On a site, install:
    - static pages plugin
    - the translate plugin
1. In the languages settings, create a second language (for example FR)
1. Create a CMS layout that defines a repeater for static pages:
```HTML
description = "test layout"

[staticPage]
useContent = 1
default = 0
==
<!DOCTYPE html>
{repeater name="pageblocks"}
    {variable name="text" type="text" label="Content"}{/variable}
{/repeater}
<html>
    <head></head>
    <body>
            {% for item in pageblocks %}
            	<div>{{ item.text }}</div>
            {% endfor %}
        </div>
    </body>
</html>
```
4. Create a static page that uses this layout and fill some data inside the repeater for both languages.
4. Save the static page and refresh the browser page.
4. Open the static page again and try to switch the language of the repeater.
4. Now you can notice that:
    - The language indicator of the repeater is correctly displaying the wanted language
    - But the values displayed are the one from the default language
    - If you look at the values of the static page using a text editor, you can see that the values are still correct.
    - If you save the static page from the backend (after having tried to switch locale), and reopen the translated static page file with your text editor, you will see that the values have been replaced by the values of the default locale.

#### October build
430

#### A possible solution
[...]

Issue #290, opened 2017-08-12
MLRepeater not working with groups


munxar:
Expected behavior

Add a group type X in language A and a group type Y in language B. If I switch from language A to B and back to A, the group type should be X.

Actual behavior

If the language is switched from B to A the group type is Y instead of X.
Check Video: [http://youtu.be/a633InT4pwo?hd=1](http://youtu.be/a633InT4pwo?hd=1)
(ignore audio...)

Reproduce steps

1. DB table with text columns id (PK),  content (text) 
2. Model from the table, make 'content' jsonable
3. Define columns.yaml (not important), 
4. Define fields.yaml like this:
```yaml
fields:    
    content:
        label: Content
        oc.commentPosition: ''
        prompt: 'Add new item'
        span: full
        type: repeater
        groups:
          textarea:
            name: Textarea
            description: Textarea
            icon: icon-leaf
            fields:
              textarea_content:
                label: Content
                type: richeditor
          text:
            name: Text
            description: Text
            icon: icon-leaf
            fields:
              text_content:
                label: Content
```
5. Make controller and backend menu
6. Create two languages (e.g. en (default), de)
7. Place a text in language en and set some value, place a textarea in language de and set some value -> save
8. Refresh form, switch from language en to de (group type in de is correct -> textarea), switch back to en (group type is wrong -> textarea instead of text)

October build

(419, 420)

Note: If alle languages have the same order and types of groups, everything is working as expected.

munxar:
I investigated this issue and found out that in file modules/backend/formwidgets/Repeater.php the function
`
/**
     * Returns a field group code from its index.
     * @param $index int
     * @return string
     */
    public function getGroupCodeFromIndex($index)
    {
        return array_get($this->indexMeta, $index.'.groupCode');
    }
`
always returns the group code of the default locale.

LukeTowers:
@munxar are you able to come up with a solution to fix it and provide a PR?

massimomegistus:
Hi guys,
i've made some tests in static pages with a group repeater and i've just noticed that removing the $this->getGroupCodeFromIndex call in processSaveValue, the group translation seems to work correctly, even if all languages don't have the same order and types of groups.

I have redefined processSaveValue in rainlab/translate/formwidgets/MLRepeater.php

protected function processSaveValue($value)
    {
        if (!is_array($value) || !$value) {
            return $value;
        }

        **// if ($this->useGroups) {
        //     foreach ($value as $index => &$data) {
[...]

PR #324, opened 2017-12-16
fix MLRepeater bug when switching locales

munxar:


Fl0Cri:
This fix works well when used from static pages plugin.

To test more complex repeaters, we need to use define them in a yaml file. So I did some tests using the [test plugin](https://github.com/daftspunk/oc-test-plugin). The **Countries** section have some repeaters, but these are not translatable out of box. To reproduce my tests, you will need to add the following class properties in [/plugins/october/test/models/Country.php](https://github.com/daftspunk/oc-test-plugin/blob/master/models/Country.php#L11):
```PHP
public $implement = ['@RainLab.Translate.Behaviors.TranslatableModel'];
public $translatable = ['locations', 'states', 'content'];
```

Note that I didn't test everything without the fix to compare results. I only tested when it made sense to me. In any case the "without fix" behaviour is to load the values from the default lang in place of the translated ones as described in #323.

The following schema is repeated to test each repeater type:
1. Create a fresh country and save it
1. From default locale, insert some values
1. Switch to a secondary locale
1. Insert some values
1. Save, switch back to the default locale and check the loaded values
1. Reload the page and check the values
1. Switch to the secondary locale and check the values
1. Switch bacl to the default locale and check the values

### Classic repeaters
> Found in the "Related" tab of a "Country" item, it looks like what we can do using the static pages plugin. Note that the repeater contains 2 dropdowns, with a specificity being that the second one depends on the first one.
1. Create Country - *ok*
1. Insert values - *ok* (changing the value of the first dropdown correctly affect the available options in the second one)
1. Switch locale - *ok* (repeater is cleared)
1. Insert values - *ok*
1. Save and switch locale - *ok* (values from the default locale are displayed)
1. Reload - *ok* (values from the default locale are displayed)
1. Switch locale - *ok*
1. Switch locale back to default - *ok*

In this classic use case, everything works fine. 

### Nested repeaters
> Found in the "States" tab of a "Country" item. Basically it's a repeater with each node having a text field and a repeater (sub repeater nodes having only a text field).

For this test I try to insert the following values:
```
Test [lang]
    |---- SubTest [lang]

Test2 [lang]
```

1. Create Country - *ok*
1. Insert values - *ok*
1. Switch locale - *ok*
1. Insert values - *ok*
1. Save & switch locale - *ok*
1. Reload - *ok*
1. Switch locale - **problem:** now inside `Test2 FR`, there is an empty text field that's not supposed to be here.
1. Switch locale back to default - **the empty text field is also here now**

Note: I tested the 2 last points without the fix, and the problem of an emty sub field being created is also present.
So this bug was **not** introduced by this fix. I will open an other issue, but I won't have time to investigate on it for now.

### Grouped repeaters
> Found in the "Content" tab of a "Country" item

For this test I try to insert a text content and an image (with different values and order for different languages)

1. Create Country - *ok*
1. Insert values - *ok*
1. Switch locale - *ok*
1. Insert values - *ok*
1. Save & switch locale - *ok*
1. Reload - *ok*
1. Switch locale - *ok*
1. Switch locale back to default - **problem:** the english version is empty

Note: I found after testing that it's exactly the problem pointed out by @munxar in https://github.com/rainlab/translate-plugin/issues/290: it is because in my test i had the textarea first and the image after in EN version, and the image first and textarea after in FR. If I use the same order in both language, the test is ok.

munxar:
@Fl0Cri Yes, exactly. To fix the repeaters with groups or repeater in repeater, which is basically the same, we need another fix. But this PR at least makes the repeater work again for average cases.
Btw. Thank you very much for the detailed testing!

jan-vince:
Hi, for me this fix works for Static pages and plugin form fields standard repeaters.

Would be great if this can be merged :)

LukeTowers:
@Fl0Cri @munxar @jan-vince final verdict is that this fix works and doesn't break anything else?

jan-vince:
@LukeTowers I have been testing this with and without Translate plugin and everything looks OK.

Without this fix I cannot use repeater with more than one language.

*All other translatable form fields (like text, richeditor) works form me as before.*

Proof of this working:

october-fix-mlrepeater

Fl0Cri:
@LukeTowers I think it's safe to merge. Even if it breaks something that we didn't see, it will only impact this isolated part of the code (the onSwitchItemLocale ajax handler), and this feature actually doen't work at all ;-)

LukeTowers:
Perfect, thanks for all the work on this @Fl0Cri @munxar @jan-vince!

PR #329, opened 2018-01-10
Fix config reference

acasar:
Fixes config reference as discussed here: https://github.com/rainlab/translate-plugin/pull/318

cc @LukeTowers @vannut


vannut:
Super @acasar 
Would you mind adding a new version in `updates/version.yaml` so that the update is propagated to already deployed plugins?

LukeTowers:
@acasar if you could add a new entry to version.yaml and check the rest of the plugin for any other broken config references I'll merge this

acasar:
@LukeTowers @vannut Thanks for feedback. I updated the version and also check all other config references, but this one was the only one without a prefix.

PR #332, opened 2018-01-15
Fix repeater content not updated on switch locale

trnikon:
On the back end when switching locale to a repeater with a rich editor the field contents would not be updated and the default language contents (en) would be shown.


trnikon:
Was already fixed

PR #335, opened 2018-01-21
Configurable cache timeout. Resolves #334

### LINKED ISSUE #334: Cache timeout customization with env
Hello,

Currently cache timeout is hard coded to 1440. 

There are some occasions though - for example static websites where content or translations do not change, where it would make sense to increase that timeout. Otherwise every 24h some unlucky visitor, or a robot, will trigger a lot of queries, which can convert to several seconds of additional initial request time, because queries are not collected by run one after another. 

My suggestion would be to add possibility to use env and implement some simple solution like:

```
    public static function saveToCache()
    {
        if (!self::$hasNew || !self::$url || !self::$locale) {
            return;
        }

        Cache::put(self::makeCacheKey(), self::$cache, env('TRANSLATE_CACHE_TIMEOUT',1440));
    }
```

would you accept pull request with this, or maybe have some other idea to resolve it?

viamage:
Using `Config` as per @LukeTowers advise.

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -11,6 +11,7 @@ return [
     | ( For now only used in the alternate-hreflang cmponent )
     |
     */
-    'prefixDefaultLocale' => true
+    'prefixDefaultLocale' => true,
+    'cacheTimeout'        => 1440,
 
 ];
--- a/config/config.php
+++ b/config/config.php
@@ -3,15 +3,27 @@
 return [
     /*
     |--------------------------------------------------------------------------
-    | prefix the default Locale
+    | Prefix the default Locale
     |--------------------------------------------------------------------------
     |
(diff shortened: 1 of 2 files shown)


config/config.php
```diff
@@ -11,6 +11,7 @@
     | ( For now only used in the alternate-hreflang cmponent )
     |
     */
-    'prefixDefaultLocale' => true
+    'prefixDefaultLocale' => true,
+    'cacheTimeout'        => 1440,
```
LukeTowers: You need to add a descriptive comment detailing what this setting is, why someone might want to change it, and what the possible values are before I can merge this.

viamage:
Comment added

Issue #336, opened 2018-01-24
Losing URL parameters after switching locale


abdulla-allaith:
Hello,
        
I have a page that has URL parameters          
e.g. `/en/test-page?id=1`          
           
The layout supports the `[localePicker]` component and the switch language button is implemented  as documented: `<a data-request-data="locale: 'de'" data-request="onSwitchLocale" href="#">`         
          
However, when the page switches locale, the params are gone and the url becomes the following: `/de/test-page`         
        
This is very problematic because the page's content depend entirely on the parameter(s) to fetch the correct data. I have searched for similar issues but I have not found any.

LukeTowers:
@abdulla-allaith see if you can find the issue and or a fix for it in https://github.com/rainlab/translate-plugin/blob/master/components/LocalePicker.php#L69 and then make a PR to fix it please :)

abdulla-allaith:
@LukeTowers after some debugging, I have not been able to fix it myself, but I have some findings that I can share:         
          
I discovered that the parameters are always empty in 
https://github.com/rainlab/translate-plugin/blob/master/components/LocalePicker.php#L140
```
$params = $this->getRouter()->getParameters();
```
The strange thing is, when the `onSwitchLocale` is called, I tried calling:
```
\URL::current()
```
and the result is a URL without parameter(s)!            
However:
```
\URL::previous()
```
returns the correct URL which has the parameter(s).        
This is an Ajax request though, how can it be possible that the page is reloading itself before calling the Ajax? Does this happen on purpose?

munxar:
@abdulla-allaith @LukeTowers I'll already fixed that in my locale picker, happy to make a PR.
Should I add an property 'preserve query' or just appen if a query string is present?

abdulla-allaith:
@munxar Thanks and please make the PR. I believe the property would be very useful, but the default should be preserving the query.

munxar:
here you go: https://github.com/rainlab/translate-plugin/pull/337

LukeTowers:
@munxar it makes the most sense to preserve the query by default. If an issue arises that requires preserving the query to be configurable, then we can discuss that when it happens.

anastasiauz:
I have a similar problem, with up to date Translate plugin 1.4.1

If the url doesn't contain active locale segment the link is updated. After update get parameters are lost.
Ex.: link 
http://kompasgid.beget.tech/knigi?tags=skazki 
the link is updated to contain 'ru' segment -> 
[...]

PR #337, opened 2018-01-25
preserve query string, when switching locales

munxar:
should fix https://github.com/rainlab/translate-plugin/issues/336


LukeTowers:
@munxar have you tested that it does? @abdulla-allaith could you also test this and make sure that it works for you as well?

abdulla-allaith:
@LukeTowers yes I have tested this and I can confirm it is working perfectly 👌

Issue #338, opened 2018-01-26
Dates translation not working anymore


amdad:
Carbon do not react on language switch. Always show dates in default language of config/app. 
Tested on many plugins. Rainlab Blog also.

munxar:
@amdad afaik there was never a translation of dates out of the box. we had to add the twig extensions plugin to make it work.

amdad:
So, you know it wrong. It works flawlessly some times ago.

https://github.com/octobercms/october/issues/1968

amdad:
Thing is, that translation is done correctly in system by carbon extension. If you set global app config on other then English you can see it (translated dates). It just not switch to active lang with languageSwitcher like it was before last updates. 

Without latest pending 'edge' fix language switcher doesn't work switch even content properly. So latest issues must be related to this issue also.

vannut:
There is a difference between the app locale (set in the config file) and the locale which is selected by the translate plugin. 

I've overcome this 'issue' and the problem of different system locale (like nl_NL.UTF8 vs nl_NL vs nl_NL.UTF-8) by wrapping the translation of dates via carbon in a twig-function which standardises the locale-array to work on every system, takes a format as argument with a sensible (for us) default.

munxar:
@vannut was about to write the same :-)

amdad:
@vannut @munxar  Guys, I told you, this function was added by my request and works by itself. And still works. It is just broken now. 

Originally dates show up only in English. From this octobercms/october#1968 October got function to extend Carbon with localized/translated version. It still there, just somewhere between communication between Translate plugin and October something is missing after last updates. It's need to be fixed, not talking about workarounds.

'locale' => 'fr',

en, ru, fr and so on changing output of dates. In Translate plugin language settings I also from beginning have short versions of lang names. So it was working fine and should be still. Don't overthink it.

LukeTowers:
@amdad you seem to be the most knowledgeable about the issue and what broke. Please submit a detailed pull request to fix it.

amdad:
If I could, I would Luke. I just know how it works before is stops. Because I just used it.

LukeTowers:
@amdad don't be afraid to dig into the code and read / play around with it. I can tell you right now that I personally don't have time to work on this particular issue so any fix will have to come from the community.

munxar:
@amdad @LukeTowers I found the code and your totally right, Argon was added exactly for this purpose.
I'll take care of this, if nobody minds.

[...]

PR #339, opened 2018-01-28
fix date translation

munxar:
This fixes bug https://github.com/rainlab/translate-plugin/issues/338


LukeTowers:
@munxar this results because of https://github.com/laravel/framework/commit/3385fdc0f8e4890ab57261755bcbbf79f9ec828d#diff-7b18a52eceff5eb716c1de268e98d55dR1045 added in Laravel 5.4. A more robust fix will have to be made in the octobercms/library repository and we need to evaluate if any more events have been modified accordingly.

LukeTowers:
Other changed events listed here: https://laravel.com/docs/5.4/upgrade. @daftspunk did you handle this in the upgrade?

daftspunk:
Fixed in core (https://github.com/octobercms/library/commit/16f7a286d8c90fb114d3abc9f7a51218c414c1f7)

This will be available in build 433+

amdad:
"legacy" will stay forever in core? So this PR fix for Translate plugin will not be needed, ever since 433?

PR #341, opened 2018-01-31
Translate URL parameters

datune:
Added an event which triggers on language switch to allow for parameter manipulation.

Credit to @mariavilaro, I added some documentation and cleaned it a bit, so we can get this merged ;-)

@LukeTowers

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,7 +194,7 @@ Event::listen('translate.params.translateParams', function($page, $params, $oldL
         return YourModel::translateParams($params, $oldLocale, $newLocale);
     }
 });
-
+```
 
 In YourModel, one possible implementation might look like this:
 
--- a/README.md
+++ b/README.md
@@ -186,10 +186,10 @@ The word "Contact" in French is the same so a translated URL is not given, or ne
 
 ## URL Parameter translation
 
-It's possible to translate URL parameters by listening to the translate.params.translateParams event, which is fired when switching languages.
+It's possible to translate URL parameters by listening to the translate.localePicker.translateParams event, which is fired when switching languages.
 
 ```php
-Event::listen('translate.params.translateParams', function($page, $params, $oldLocale, $newLocale) {
+Event::listen('translate.localePicker.translateParams', function($page, $params, $oldLocale, $newLocale) {
     if ($page->baseFileName == 'your-page-filename') {
         return YourModel::translateParams($params, $oldLocale, $newLocale);
     }
--- a/README.md
+++ b/README.md
@@ -190,7 +190,7 @@ It's possible to translate URL parameters by listening to the translate.params.t
 
 ```php
 Event::listen('translate.params.translateParams', function($page, $params, $oldLocale, $newLocale) {
-    if ($page == 'your-page-filename') {
+    if ($page->baseFileName == 'your-page-filename') {
         return YourModel::translateParams($params, $oldLocale, $newLocale);
     }
 });
--- a/README.md
+++ b/README.md
@@ -184,6 +184,33 @@ The word "Contact" in French is the same so a translated URL is not given, or ne
 - /ru/контакт - Page in Russian
 - /ru/contact - 404
 
+## URL Parameter translation
+
+It's possible to translate URL parameters by listening to the translate.params.translateParams event, which is fired when switching languages.
+
+```php
+Event::listen('translate.params.translateParams', function($page, $params, $oldLocale, $newLocale) {
+    if ($page == 'your-page-filename') {
+        return YourModel::translateParams($params, $oldLocale, $newLocale);
+    }
+});
+
+
+In YourModel, one possible implementation might look like this:
+
+```php
+public static function translateParams($params, $oldLocale, $newLocale) {
+    $newParams = $params;
+    foreach ($params as $paramName => $paramValue) {
+        $records = self::transWhere($paramName, $paramValue, $oldLocale)->first();
+        if ($records) {
+            $records->translateContext($newLocale);
+            $newParams[$paramName] = $records->$paramName;
+        }
+    }
+    return $newParams;
+}
+```
 ## Conditionally extending plugins
 
 #### Models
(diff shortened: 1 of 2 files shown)


components/LocalePicker.php
```diff
@@ -72,6 +77,9 @@ public function onSwitchLocale()
             return;
         }
 
+        // Remember the current locale before switching to the requested one
+        $this->oldLocale = $this->translator->getLocale();
```
munxar: Is there a reason to store this in an instance variable? Wouldn't a locale variable be enough?

datune: @see https://github.com/datune/translate-plugin/blob/4d6d3695016d2a61dda1c777fd7db7d939c21ab6/components/LocalePicker.php#L156

components/LocalePicker.php
```diff
@@ -142,7 +150,15 @@ protected function makeLocaleUrlFromPage($locale)
             $router = new RainRouter;
 
             $params = $this->getRouter()->getParameters();
-
+            
+            // Provide an opportunity to manipulate the parameters
```
LukeTowers: Needs inline documentation for the event, see https://github.com/octobercms/october/blob/master/modules/system/console/OctoberMirror.php#L96-L108

components/LocalePicker.php
munxar: yes, I know the original PR but I think a local variable would do. Or give me an example why a public instance variable makes more sense?

datune:
@LukeTowers I added the inline documentation.

LukeTowers:
Awesome, thanks! Why are we only passing the page filename and not the entire page object?

datune:
@LukeTowers Good call, I don't see any reason why we should limit this to the baseFilename, so I changed it.

components/LocalePicker.php
```diff
@@ -142,7 +150,29 @@ protected function makeLocaleUrlFromPage($locale)
...
+                        
+            /**
+             * @event translate.params.translateParams
+             * Enables manipulating the URL parameters
+             *
+             * You will have access to the page object, the old and new locale and the URL parameters.
+             *
+             * Example usage:
+             *
+             * Event::listen('translate.params.translateParams', function($page, $params, $oldLocale, $newLocale) {
```
LukeTowers: Please add four spaces of indentation for the example code in this comment. When we get API documentation generation for plugins, this will properly format the example as a code block in the documentation

datune:
@LukeTowers Done.

README.md
```diff
@@ -184,6 +184,33 @@ The word "Contact" in French is the same so a translated URL is not given, or ne
...
+});
+```
+
+In YourModel, one possible implementation might look like this:
+
+```php
+public static function translateParams($params, $oldLocale, $newLocale) {
+    $newParams = $params;
+    foreach ($params as $paramName => $paramValue) {
+        $records = self::transWhere($paramName, $paramValue, $oldLocale)->first();
```
LukeTowers: Just thought of this now @munxar, thanks for your patience with this. Is this an optimal query? I feel like this could cause a lot of extra queries to be run, one for every URL parameter. What are your thoughts on that? It's not a deal breaker for the PR but I'd like to try and provide as good as an example as possible.

datune: @LukeTowers I think that even with 10 URL parameters it shouldn't matter, there just read queries. If performance is really an issue the implementer should cache the results. (not to mention the DB usually has a query cache anyways...)

munxar:
Agreed, and it Just happens on the switching of the locale. Nothing wrong
with the example.


Am 07.02.2018 5:08 nachm. schrieb "datune" <notifications@github.com>:

*@datune* commented on this pull request.
------------------------------

In README.md
<https://github.com/rainlab/translate-plugin/pull/341#discussion_r166666571>
:

> +
+```php
+Event::listen('translate.params.translateParams', function($page,
$params, $oldLocale, $newLocale) {
+    if ($page->baseFileName == 'your-page-filename') {
+        return YourModel::translateParams($params, $oldLocale, $newLocale);
+    }
+});
+```
+
+In YourModel, one possible implementation might look like this:
+
+```php
+public static function translateParams($params, $oldLocale, $newLocale) {
+    $newParams = $params;
+    foreach ($params as $paramName => $paramValue) {
+        $records = self::transWhere($paramName, $paramValue,
$oldLocale)->first();

@LukeTowers <https://github.com/luketowers> I think that even with 10 URL
parameters it shouldn't matter, there just read queries. If performance is
really an issue the implementer should cache the results. (not to mention
the DB usually has a query cache anyways...)

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/rainlab/translate-plugin/pull/341#discussion_r166666571>,
or mute the thread

LukeTowers:
Awesome, thanks for your work on this @datune & your reviews @munxar. I'll get this merged, but leave #244 and #227 open for potential future work on translating components.

components/LocalePicker.php
```diff
@@ -142,7 +150,29 @@ protected function makeLocaleUrlFromPage($locale)
...
+             * Example usage:
+             *
+             *     Event::listen('translate.params.translateParams', function($page, $params, $oldLocale, $newLocale) {
+             *        if ($page->baseFileName == 'your-page-filename') {
+             *             return YourModel::translateParams($params, $oldLocale, $newLocale);
+             *         }
+             *     });
+             *
+             */            
+            $translatedParams = Event::fire('translate.params.translateParams',
```
LukeTowers: One final change, please change the event name to `translate.localePicker.translateParams` everywhere. Thanks!

datune:
@LukeTowers Event name has been changed everywhere. Thanks for your feedback and support in this matter, much appreciated!

LukeTowers:
Not a problem, thank you for your contributions!

mariavilaro:
Hi! this is working super good in my projects, when will you update the build? :)

LukeTowers:
@mariavilaro do you mean version.yaml? If so, submit a PR to update it yourself with an appropriate update message.

mariavilaro:
I mean the published plugin version in octobercms plugins is still version 1.3.8

imagen

LukeTowers:
@mariavilaro done, it should be released shortly.

PR #345, opened 2018-02-05
Add operator option for TransWhere function

abdulla-allaith:
Add operator option for TransWhere function, with default option being '='.
Most likely use of this new option is to use "LIKE" operator instead of an equal value.
In order to use it, the locale variable will need to have either a locale value or `null`.

### DIFF
--- a/behaviors/TranslatableModel.php
+++ b/behaviors/TranslatableModel.php
@@ -30,7 +30,7 @@ class TranslatableModel extends TranslatableBehavior
      * @param  string $locale
      * @return Builder
      */
-    public function scopeTransWhere($query, $index, $value, $locale = null)
+    public function scopeTransWhere($query, $index, $value, $locale = null, $operator = '=')
     {
         if (!$locale) {
             $locale = $this->translatableContext;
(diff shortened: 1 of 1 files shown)


LukeTowers:
@a7madev @abdulla-allaith could you please provide some code examples as to why this is required?

abdulla-allaith:
We wanted to add a search feature with translation. The problem was it only accepting the exact phrases in the database (e.g. `transWhere("title", "Search Term")`). In order to extend the search functionality to accept partial phrases, we needed to do a `where` statement with the `LIKE` operator. This provides flexibility for other types of searches as well (e.g. `">"`, `"<="`). The new `where` statement utilizes Laravel's `where("title", "LIKE" ,"%Search%")` to find partial matches as well. However, with `transWhere` which has a `locale` parameter, to change the operator, the locale parameter has to be passed with either a value (`"en"`, `"de"`) or `null`.           
          
An actual usage example:         
`transWhere("title", "%Search%", null, "LIKE")`        
I tested this and it works as required.

LukeTowers:
Perfect, thanks for the explanation!

PR #349, opened 2018-02-09
added missing event import

munxar:
The current version is missing import for the Event facade. This PR fixes this.
@LukeTowers this is very important, because current version is broken.


LukeTowers:
Merged in https://github.com/rainlab/translate-plugin/commit/61e83fdd09fe4b51ddf0943b2e7ca5ced46e5b8a instead because this PR was too cluttered with other unrelated commits

LukeTowers:
Thanks for catching that!

Issue #350, opened 2018-02-18
date format does not respect locale


rounce:
I have a multilingual site. All is well when the user is on the English version of the site. For formatting dates, I use format and it always produces the result in English. (because 'locale' => 'en' in config/app.php)
for example
```
$translator = Translator::instance();
$translator->setLocale("ru", false);
dd($user->created_at->format("F y"));
```
results in "May 16"
When i type  'locale' => 'ru' in config/app.php
result always "май 16".

rounce:
For now it is necessary to set the locale by force:
```
$translator = Translator::instance();
$user->created_at->setLocale($translator->getLocale());
```

LukeTowers:
Update your October instance, it already did although support was temporarily broken between Build 420 & Build 433. #338

PR #351, opened 2018-02-18
 Make Date take into account the locale

rounce:
https://github.com/rainlab/translate-plugin/issues/350

As an option. I'm not sure that it will work in all cases.


LukeTowers:
Update your October instance, it already did although support was temporarily broken between Build 420 & Build 433. #338

Issue #205, opened 2016-12-25
Bug On CMS


iboldurev:
Hello!

CMS Pages

There is no way to translate the names of pages, only the translation of links to the page

How to fix? Please help

BUILD Version OctoberCMS 390
Plugin version 1.2.7 - 1.2.8

![2016-12-25 12 59 40](https://cloud.githubusercontent.com/assets/17103170/21470874/a702193c-caa3-11e6-9d77-bf7e2fdf5dfe.png)

P.S

Static pages plugin it works fine
![2016-12-25 14 12 18](https://cloud.githubusercontent.com/assets/17103170/21471079/86004e26-caac-11e6-96f8-f84e9ee8778b.png)

tbergeron:
I don't think this plugin should help translate anything on the CMS tab, only in the static pages tab. I wonder why there's a language picker un the URL field though. I have the same here right now.

Eoler:
> I wonder why there's a language picker un the URL field though.

Because the plugin actually advertises URL translations here:
https://github.com/rainlab/translate-plugin#url-translation

ctf0:
i believe this is intentional because for the pages plugin the page is saved in the db, so u have more than locale because its treated as a normal json column that gets fetched according to the current local,
however for the pages under cms, the page gets saved on the disk, so to have different names for the same file it means it have to be duplicated
as doc says
> Pages will not be duplicated for a given language

LukeTowers:
@ctf0 The pages in the static pages plugin (this) are saved to disk, not to DB

ctf0:
@LukeTowers okey, so why cant we have the same option with cms ? i mean what are the limitations ?

LukeTowers:
I'm not sure, I didn't implement the integration with the Pages plugin. @daftspunk could probably provide more insight. It may be that it simply hasn't been implemented yet, feel free to dig into the existing integration with the Pages plugin to see what you can do as far as implementing it for the CMS pages via a PR

ctf0:
i would gladly implement it if someone could give me some guide lines, am still new into this but surely would love to contribute :100:

LukeTowers:
First step would be to look for any code within this plugin that references the RainLab.Pages plugin or look in the RainLab.Pages plugin for any code that references this plugin. Then take it from there. Apologies I can't provide more guidance than that to start, I'm not hugely familiar with how it was implemented and don't have time to look into it myself right now.

ctf0:
[...]

Issue #138, opened 2016-06-09
[BOUNTY] Allow translation of CMS Page properties


pavlo-tk:
Bounty link: https://www.bountysource.com/issues/35021479-bounty-allow-translation-of-cms-page-properties

Bounty specification:
- Translate the page Description property
- Translate the page Title property
- Translate the page Meta Description property
- Translate the page Meta Title property

Discovery is needed for the best way to achieve this. There is a solution in place with no supporting UI. 

Possible options:
1. A generic key seeder can be used, similar to Builder plugin.
2. Implement a Translatable behavior that proxies the values to the page settings.

Original notes:

Hello.
In StaticPages plugin we have ability to translate pages title, meta_description, etc.
Do you have plans to add such ability to standard CMS pages to translate the same form fields (title, meta_description)?

norgen:
Maybe:
`<meta name="description" content="{{ this.page.meta_description|_ }}">
 <meta name="keywords" content="{{ this.page.meta_keywords|_ }}">
`

side-by-side:
I know about that method, but I wonder if there any plans to make this available like in Static Pages, via pages forms directly.

amdad:
Right, translations of this via Settings = terrible UX.

ilyaskorik:
+100500 to this

daftspunk:
Here's how I would do it

```
<title>{{ this.page.meta_title|_ }}</title>
<meta name="description" content="{{ this.page.meta_description|_ }}">
<meta name="keywords" content="{{ this.page.meta_keywords|_ }}">
```

Set each value to something like

- `mypage.title`
- `mypage.description`
- `mypage.keywords`

Then open `theme.yaml` and include:

```
en:
    mypage.title: "Title in English"
    mypage.description: "Description in English"
    mypage.keywrods: "keywords, in, english"

ru:
   mypage.title: ...
   ...
```

Right now there is no UI for this approach, we could make it by borrowing the on-the-fly language key creation process from the RainLab.Builder plugin.

thomasilmer:
Bounty added on BOUNTYSOURCE to implement the input locale switcher in CMS pages

mortezaataiy:
i do all this works but
not translatable in the backend edit field :(

mjauvin:
Fixed by https://github.com/rainlab/translate-plugin/pull/367

PR #363, opened 2018-03-28
Translatable page

### LINKED ISSUE #331: Meta title/description for CMS page not translatable by default
I just found out that a CMS page's Meta title and Description is not translatable by default. 
schermafbeelding 2018-01-11 om 10 30 23

Only the slug is translatable. 
Am I missing something here?
Or might it be a good thing to make those fields also translatable by default?

mjauvin:
- Add CMS Page translation (title, description, meta-title and meta-description)
- Fix a typo in "hasTranslatableAttributes" method name

Resolves #331


munxar:
@mjauvin 👍 thank you!

munxar:
@LukeTowers @mjauvin tested this feature. looks good.

classes/TranslatableBehavior.php
```diff
@@ -297,7 +297,7 @@ public function lang($context = null)
      * Checks if this model has transatable attributes.
      * @return true
      */
-    public function hasTransatableAttributes()
+    public function hasTranslatableAttributes()
```
LukeTowers: You need to update the call to the method here as well: https://github.com/rainlab/translate-plugin/blob/e603c487fa0c1c4a9ca31cd6c1a9305ea2364389/classes/EventRegistry.php#L84 do a search of the code to see if it was misspelled anywhere else too.

LukeTowers: Never mind, I see you fixed it in https://github.com/rainlab/translate-plugin/pull/363/files#diff-252810a1b0bce82b1a1ef724cee3c8deR96

LukeTowers:
@munxar could you find all the issues for this that have been posted in the octobercms/october, rainlab/pages-plugin, and rainlab/translate-plugin and link them here so that I can close all of them when I merge this?

mjauvin:
@LukeTowers Hmmm, rainlab.pages uses '@RainLab.Translate.Behaviors.TranslatablePageUrl' and I removed it... I'll add it back. Would be nice to completely remove it, though, what's the best course of action?

munxar:
@LukeTowers I think these are the related ones:
- https://github.com/rainlab/translate-plugin/issues/331
- https://github.com/rainlab/translate-plugin/issues/205
- https://github.com/rainlab/translate-plugin/issues/138

@mjauvin Imho the RainLab.Pages plugin should not depend on the translate plugin, but the other way round.
for all other controls the instantiation call comes from the translate plugin, but here (https://github.com/rainlab/pages-plugin/blob/master/classes/Content.php) the pages plugin does it.

shouldn't we move this into the pages plugin? @LukeTowers @mjauvin Wha do you think?

Btw. you could mark the TranslatableUrl as depreciated just in case someone else is using it.

updates/version.yaml
```diff
@@ -49,3 +49,4 @@
 1.3.7: Fix config reference in LocaleMiddleware
 1.3.8: Keep query string when switching locales
 1.4.0: Add importer and exporter for messages
+1.4.1: Add TranslatablePage behavior (replaces TranslatablePageUrl)
```
munxar: remove the (replace TranslatablePageUrl)

mjauvin:
@LukeTowers anything I can do to help close this PR?

updates/version.yaml
```diff
@@ -49,3 +49,6 @@
 1.3.7: Fix config reference in LocaleMiddleware
 1.3.8: Keep query string when switching locales
 1.4.0: Add importer and exporter for messages
+1.4.1:
```
LukeTowers: Is this form of release notes actually supported by October? Typically multi-line versions are to provide one comment about the version and any DB migration scripts to run.

classes/EventRegistry.php
```diff
@@ -29,25 +29,36 @@ public function registerFormFieldReplacements($widget)
...
         }
 
-        if (
-            $model instanceof Page &&
-            isset($widget->fields['settings[url]'])
-        ) {
-            $widget->fields['settings[url]']['type'] = 'mltext';
+        if ($model instanceof Page)
+        {
+            if (isset($widget->fields['settings[title]']))
```
LukeTowers: Is there a cleaner way to do this? Why can't we use https://github.com/rainlab/translate-plugin/blob/master/classes/EventRegistry.php#L107?

updates/version.yaml
mjauvin: You may be right, I'll change this.

mjauvin:
Closed and moved to #367

PR #364, opened 2018-03-29
Preserve any query string parameters when forcing the url schema

tobias-kuendig:
If a user visits a page `/search?q=foo` while the `forceUrl` property is set to `true`
a redirect to `/en/search` is returned and the query parameters get lost.
With this change all query parameters from before the rewrite are added back
to the redirect url.


munxar:
An event like for the params would be cool, just in case someone wants to translate the query string keys.

tobias-kuendig:
Great idea! I've added the event.

components/LocalePicker.php
```diff
@@ -179,4 +177,40 @@ protected function makeLocaleUrlFromPage($locale)
...
+         *        if ($page->baseFileName == 'your-page-filename') {
+         *             return YourModel::translateParams($params, $oldLocale, $newLocale);
+         *         }
+         *     });
+         *
+         */            
+        $page = $this->getPage();
+        $query = request()->query();
+
+        $translatedQuery = Event::fire('translate.localePicker.translateQuery',
```
LukeTowers: Event documentation needs to occur immediately before the firing of the event itself

LukeTowers: This event should also be documented in the README

tobias-kuendig: I have added a paragraph to the README and moved the comment. :+1:

LukeTowers:
@tobias-kuendig @munxar, would it make more sense to just have this fire the `translate.localePicker.translateParams` event instead of having its own event or does it make more sense for it to be its own event?

munxar:
@LukeTowers I had the same idea, but then if I listen on the event I don't know where the param is from. Imagine the (rare) case where you have a query string and a url param with the same name (e.g. 'id').
So imho it makes sense to have a own event. Isn't there a wildcard thingy like 'translate.locale.*' in laravel events? so if someone needs both, use this. no?

LukeTowers:
@munxar good point.

hlev:
I'm glad I looked at the PRs before submitting an issue. Is this moving forward? If it's held up, is it OK to patch the file in the deployed October application or would it mess up verification of file hashes (if any)?

LukeTowers:
@hlev you're probably fine patching in the meantime, just make sure to check in when this gets merged to make sure your code matches the final version. @tobias-kuendig @munxar what else was remaining on this?

hlev:
Thanks @LukeTowers, did just that and all is well.

tobias-kuendig:
What's the state of this PR? Are any futher changes needed?

w20k:
@tobias-kuendig you could ask someone to re-test :) And that's would be all.

mjauvin:
I have tested this PR and it's working as expected.

@bennothommo what MileStone should we set this to?

mjauvin:
Anybody else can test this on their end? @munxar @osmanzeki @mariavilaro ?

bennothommo:
@mjauvin it doesn't look like a breaking change to me so 1.6.1 should be fine.

PR #372, opened 2018-05-10
Fix issue where default texts are overwritten by import

acasar:
When importing translated messages, default language (denoted by x) is currently overwritten with codes. Example:

Before import:

- Code = test.message 
- Default = Test Message

After import:

- Code = test.message
- Default = **test.message**

This PR fixes that issue, so that only translations are imported. Default language (usually coming directly from the template source code) is not changed.

### DIFF
--- a/models/MessageImport.php
+++ b/models/MessageImport.php
@@ -43,6 +43,10 @@ class MessageImport extends ImportModel
 
                     $message->message_data = array_merge($message->message_data, $result);
 
+                    if(!isset($message->message_data[Message::DEFAULT_LOCALE])) {
+                        $result[Message::DEFAULT_LOCALE] = $code;
+                    }
+
                     if ($message->exists) {
                         $this->logUpdated();
                     } else {
--- a/models/MessageImport.php
+++ b/models/MessageImport.php
@@ -35,7 +35,6 @@ class MessageImport extends ImportModel
                     $code = $result[$codeName];
                     // modify result to match the expected message_data schema
                     unset($result[$codeName]);
-                    $result[Message::DEFAULT_LOCALE] = $code;
 
                     $message = Message::firstOrNew(['code' => $code]);
 


LukeTowers:
@munxar @vannut could you two take a look at this? What was the original rational behind having that line included?

vannut:
I'm not the author of that line :) So I cant help you with the rational behind it :)

munxar:
@LukeTowers @acasar It is the behavior if a message is parsed from e.g. a template. 
But as noted, it is stored in the 'x' key and if you use your real default language e.g. 'en' it should take that value. Imho it's just to display the codes in the message table.
I have no idea what side effect it has to remove this 'x' key.

acasar:
@munxar since we are doing array_merge with original values https://github.com/rainlab/translate-plugin/blob/master/models/MessageImport.php#L45 this change doesn't remove the "x" key. It just leaves it as it was before the import, which I believe correct.

My problem is that the "x" key is visible to the client as a default text and I don't want it to be replaced by codes as it would just confuse the client.

munxar:
@acasar What if the the key is new? It will not get the default value set, right?
prop. we have to distinguish between create vs. update?

acasar:
@munxar Hm.. What is the use case of adding new translations to the import file? I mean new texts won't be visible anywhere until they are manually added to the template. And even then they can collide with the "Scan for messages" functionality that may produce a different message code. But still, you are correct, that case is not handled in my PR.

Even so I still think it's wrong to save a code as a default language since it can actually be displayed on the front-end if no translation is available. So we should fill it with an actual default language text, not a code.

munxar:
@acasar I did not implement the logic with the codes, I just tried to mime the existing behavior as close as possible.
The use cases for creating translations are 
a) you can give that csv to an external translation office
b) recreate a blank october instance with your translations (multi stage setup, or local development).

acasar:
@munxar but when doing an a) use case, the codes are overwritten with "gibberish" as my client put it :) And there is currently no way to prevent that. (this is an actual complaint I got from the client).

So I think we have an option to only "create" a default language when creating a new translation. If Default language already exist, we leave it as it is. Is that ok with you?

munxar:
yes, that was my first proposal.

acasar:
Done

acasar:
@LukeTowers Can this be merged? It's been a few months wince it was confirmed.

LukeTowers:
@acasar sure, if I don't do anything about an issue it's usually because I've forgotten about it. I work pretty much exclusively off of my GitHub notifications list because there isn't enough time for me to go through and review the lists of open issues and PRs to check the status; so if there's ever anything you want some more action just ping me and I'll check in. 

If my [Patreon](https://www.patreon.com/luketowers) ever takes off then I'll be able to spend more time going through the open issues & PRs, until then _the squeaky wheel gets the grease_ 😄

acasar:
Thanks @LukeTowers 
:+1: :smiley:

PR #398, opened 2018-08-14
Detect language from browser accept languages

mauserrifle:


classes/Translator.php
```diff
@@ -136,6 +136,23 @@ public function loadLocaleFromRequest()
...
 
+    /**
+     * Sets the locale based on the browser accept languages
+     * @return bool
+     */
+    public function loadLocaleFromRequestLanguages()
+    {
+        $locales = Locale::listAvailable();
+
+        foreach(Request::getLanguages() as $request_lang) {
```
LukeTowers: add spaces after the foreach and if please, `$request_lang` to `$requestLang` please. Thanks!

LukeTowers:
@mauserrifle Awesome, thanks! Just waiting on a couple of minor changes.

LukeTowers:
Awesome @mauserrifle! If you can get another person to test and verify that it works then I'll merge. Perhaps @datune or @vannut?

datune:
@mauserrifle @LukeTowers @vannut 

I tend to disagree with this approach. It makes no sense to couple the detection of the locale sent from the browser based on whether or not I want the default locale prefixed in the URL. 

Detecting browser language preferences should work indepenently of whether or not I want the default locale prefixed. 

I'd much rather see an additional config setting allowing us to turn this feature on or off.

I took a stab at it, but it's not working with prefixDefaultLocale set to off. I'm not sure why. Maybe @vannut can come to the rescue?

```php
<?php namespace RainLab\Translate\Classes;

use RainLab\Translate\Classes\Translator;
use Closure;
use Config;

class LocaleMiddleware
{
    /**
     * Handle an incoming request.
     *
     * @param  \Illuminate\Http\Request  $request
     * @param  \Closure  $next
     * @return mixed
     */
    public function handle($request, Closure $next)
    {
        $translator = Translator::instance();
        $translator->isConfigured();

        if (!$translator->loadLocaleFromRequest()) {
            if (Config::get('rainlab.translate::prefixDefaultLocale')) {
                if (!$translator->loadLocaleFromSession()) {
                    $this->loadLocaleFromRequestLanguages($translator);
                }
            }
            else if (!$this->loadLocaleFromRequestLanguages($translator)) {
                $translator->setLocale($translator->getDefaultLocale());
            }
        }

        return $next($request);
    }

    /*
     * Set Locale based on the accepted Request (usually browser) languages
     *
     * @return bool
     */
    protected function loadLocaleFromRequestLanguages($translator)
    {
        if (Config::get('rainlab.translate::setLocaleFromRequestLanguages')) {
            return $translator->loadLocaleFromRequestLanguages();
        }

        return false;
    }
}

```

vannut:
> ...we would be able to add aliases to our defined language codes in the backend...

Maybe that's not a bad idea: adding aliasses to a certain 'cms-locale' with a fallback to the default.

Samuell1:
This plugin mostly does what you need guys:

https://github.com/Excodus/oc-translate-extended/blob/master/routes.php#L42

mjauvin:

PR #413, opened 2018-10-09
Explain how to make default translation effective

pikanji:
Added explanation to README.md how to make default translation added in theme.yaml effective. I had to dig into the source code to figure it out. I wonder why no one has asked about this, though.


LukeTowers:
Awesome, thanks!

PR #416, opened 2018-10-19
Improve implementation of MLControl

LukeTowers:
I made these changes in April and never committed them, can't remember if it worked or not. I feel like it did but no idea now.

### DIFF
--- a/formwidgets/MLMarkdownEditor.php
+++ b/formwidgets/MLMarkdownEditor.php
@@ -78,18 +78,20 @@ class MLMarkdownEditor extends MarkdownEditor
         }
     }
 
-    protected function actAsParent($switch = true)
+    /**
+     * {@inheritDoc}
+     */
+    protected function getParentViewPath()
     {
-        if ($switch) {
-            $this->originalAssetPath = $this->assetPath;
-            $this->originalViewPath = $this->viewPath;
-            $this->assetPath = '/modules/backend/formwidgets/markdowneditor/assets';
-            $this->viewPath = base_path().'/modules/backend/formwidgets/markdowneditor/partials';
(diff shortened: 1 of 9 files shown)


LukeTowers:
@datune @vannut @jimcottrell @Fl0Cri @ChVuagniaux @pvullioud 
Could you guys test this with a fresh install of October and RainLab.Translate and let me know if this magically solves the problems there were with nested repeaters? I made these changes April 26th and never actually committed them so I have no idea if I was on the right track or not.

Fl0Cri:
I could test this on a clean install with just "static pages" and "translate" (using the same static pages layout as here https://github.com/rainlab/translate-plugin/pull/409#issuecomment-425029304 for testing repeater groups) unfortunately it doesn't work:

Tested the base case, as described here https://github.com/rainlab/translate-plugin/issues/307#issuecomment-353291488 (but this time with groups, it would probably do the same with standard repeaters).
It doesn't work, I get the same results as in the second screenshot with empty fields appearing inside a normally empty repeater when switching locale.

Then I also tried this https://github.com/rainlab/translate-plugin/issues/307#issuecomment-384177728 with same result (no improvement) for both behaviors.

Finally I tried to reproduce both problems I noticed here https://github.com/rainlab/translate-plugin/pull/409#issuecomment-423906767.
The first is not reproductible because of the main bug (lot of blank subitems being created).
I could reproduce the second (only with repeater groups).

LukeTowers:
On second review, I think this was actually to supporting adding a new ML formwidget, maybe for the mediafinder?

LukeTowers:
Yup, this is in fact to support adding a new ML formwidget.

PR #419, opened 2018-11-06
Add ability for models to provide computed data

foxik0169:
Hello,

I've lately come across an undesired behaviour with blog plugin (https://github.com/rainlab/blog-plugin/issues/299). As I've explained in the issue, the field content_html is "computed" from the content field in the beforeSave function and the translated value **does not get updated**.

This, however, is true only if you don't use the original MLBlogMarkdown control because it currently has the logic of computing content_html field implemented internally. 

I've tried to come up with the most reasonable solution and my best implementation so far would be to fire an event, before saving translation data, that can provide computed values which then would be added to the translated attributes array.

The counterpart to this implementation in subclasses of Model would look like this:

```
public function __construct(array $attributes = [])
{
    parent::__construct($attributes);

    $this->bindEvent('model.translate.resolveComputedFields', function ($locale) {
        return [
            'content_html' =>
                self::formatHtml(
                    $this->asExtension('TranslatableModel')
                        ->getAttributeTranslated('content', $locale))
        ];
    });
}
```

Thanks for consideration and feedback.
Daniel

### DIFF
--- a/behaviors/TranslatableModel.php
+++ b/behaviors/TranslatableModel.php
@@ -84,8 +84,27 @@ class TranslatableModel extends TranslatableBehavior
             return;
         }
 
-        /*
-         * Resolve computed fields before saving...
+        /** 
+         * @event model.translate.resolveComputedFields
+         * Resolve computed fields before saving
+         *
+         * Example usage:
+         *
+         * Override Model's __construct method
+         *
+         * public function __construct(array $attributes = [])
+         * {
+         *     parent::__construct($attributes);
+         *
+         *     $this->bindEvent('model.translate.resolveComputedFields', function ($locale) {
+         *         return [
+         *             'content_html' =>
+         *                 self::formatHtml($this->asExtension('TranslatableModel')
+         *                     ->getAttributeTranslated('content', $locale))
+         *         ];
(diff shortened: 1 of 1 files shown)


mjauvin:
@foxik0169 so what you're saying is this is required if a plugin replaces the default BlogMarkdown formwidget and no special code is used by this plugin to handle the translation for the computed field?

mjauvin:
Ok, I was able to reproduce this and applying your PR fixed the issue.

If you can add the example on how to hook into this event (as you wrote above in this PR) right before the event fires, I'll consider merging this.

Thanks.

behaviors/TranslatableModel.php
```diff
@@ -84,6 +84,13 @@ protected function storeTranslatableData($locale = null)
             return;
         }
```
mjauvin: insert usage example right here

mjauvin:
Fixes https://github.com/rainlab/blog-plugin/issues/299

mmorells:
Still having this issue in version 1.6.2

w20k:
@mmorells open a new issue and add more details about your issue, and steps to reproduce.

Issue #256, opened 2017-05-07
Message::makeMessageCode() collisions


matteotrubini:
In italian (and maybe in other languages) we use the same phrase construction to make a question, simply adding a "?" at the end.

**Reproduce steps:**
{{ ''Beer"|_ }}
{{ ''Beer?"|_ }}

**Expected behavior:**
Beer
Beer?

**Actual behavior:**
Beer
Beer

I just tried to switch the return value to sha1($messageId) and it seems to be a good replacement that assure 1:1 string matching without any other change. This would prevent "collision" also in case of >250 char texts.

If this change is acceptable I think that should be easy ensure BC simply updating "code" column during migration.

daftspunk:
Use codes instead
```
{{ ''beer"|_ }}
{{ ''beer_alt"|_ }}
```

matteotrubini:
I got it, even if it could be a "problem" developing themes where we need to place keys instead of main languages' texts, which allows you to have themes more readable, like if you use something like gettext().
It could be an easier approach instead of use multiple content files. If you change a text in theme you MUST update according the translations, leaving blank the column of the source language.

Can hashes and "readable" codes coexist?

PR #422, opened 2018-11-15
Missing DEFAULT_LOCALE export

matteotrubini:


LukeTowers:
@matteotrubini reason for change?

munxar:
@matteotrubini I think there is a miss understanding how this 'default' locale works.
Let's assume you have two Locales configured *en* and *de* where *en* is default.

If you import messages like this `{{ 'test' | _ }}`, you will end up with an entry in the messages table that looks like this:

| id | code | message_data |
| - | - | - |
| 1 | test | {"x":"test"} |

That means the *default* matches the *code*.
now let's translate it in the backend ui with the following values for *en* and *de*:

| English (default) | English | German | 
| - | - | - |
| test | This is a test. | Das ist ein Test. |

Note how the column *English (default)* can't be changed in the backend, and is the same value as the *code*.

The Database will look like this after updating the values:

| id | code | message_data |
|-|-|-|
| 1 | test | {"x":"test","en":"This is a test.","de":"Das ist ein Test."} |

And this is it, no way to change the "x" to a different value. It will be same as the *code*.
To be honest I have no clue why this is done, because it is redundant with the *code* column, but the message table in the backend uses this values to display the **non editable** column *English (default)*.

So what will happen if you change the *x* default values to let's say **foo*? Well you'll end up with with this:

| English (default) | English | German | 
| - | - | - |
| foo | This is a test. | Das ist ein Test. |

The translation will still work the same, but your user will end up loosing the link to the *code*.

This is why I set the default to the codes value in the importer, which reflects the current behavior.
I hope this clears things up. I'm happy to discuss if this "x" inside of the message_data makes sense at all, because it now reflects only the code value and can't be changed.

probably @daftspunk can give more insight here?

matteotrubini:
Preamble: I work developing websites for tourism's businesses from 16 years, so multilanguage websites are daily tasks and over the time we have been developed some practices to speedup translations manage.

> That means the default matches the code.

Not in all cases. If I import `{{ 'This is a test.'|_ }}` I have this table:

| id | code | message_data |
| --- | --- | --- |
| 1 | this.is.a.test | {"x":"This is a test."} |

IMHO there's a wrong approach about `code`. It simply should not exists or to be changed to sha1 hash to use it as query unique index.

I used Yii years ago and from it I mutuate the translation business logic: `code` does not exists as all text are translated directly from case sensitive source string, like in `gettext()`.
Also Laravel expect "Translation Strings As Keys", because short string in some case can confuse.

This logic will solve problems like "Beer" and "Beer?" and permit to code the source files in default language (it's a tremendous speedup) ensuring coherence over the time. Currently when I have a long paragraph (>250) it's code will be "truncated" and then if I change a word at the end (after 250 chars) the code will not change, potentially invalidating the translation.

Backward compatibiliy can be ensured during plugin update at migration time.

Some reference:
https://github.com/rainlab/translate-plugin/issues/256
https://laravel.com/docs/5.5/localization#using-translation-strings-as-keys
https://www.yiiframework.com/doc/guide/1.1/en/topics.i18n#message-translation

In the meanwhile, to cover my needs I created a custom plugin to workaround translation table and use php files formatted as below:
```
return [
    'This is a test.' => 'Questo è un test.',
    ...
];
```
and works like a charme.

Unfortunately currently I haven't the time to create a fork, but I can sponsorship some PR directly or via Bountysource, if anyone it's interested.

> English (default)	English	German
> test	This is a test.	Das ist ein Test.
> Note how the column _English (default)_ can't be changed in the backend, and is the same value as the _code_.

This create confusion: which is the difference between `English (default)` and `English`? At least `English (default)` column should be renamed to `Code`, `Source` or something else.

> To clear up your points:
> 
> 1. no, it's not expecting it, it set's it to the codes value

Only `if !isset($message->message_data[Message::DEFAULT_LOCALE])` in source file, otherwise it will be imported.

> 2. you can, I do it all the time (code + all locales is all you need)

If `DEFAULT_LOCALE (code)` is not exported and I have used source string as key, without "translate" it to the real default locale (like "en"), I need to get it from db.

> 3. yes, me too and that is what you can do right now.

Same as 2.

munxar:
@matteotrubini thank you very much for clarifying the *code* vs *default*. When I developed the export/import feature I was simply not aware of this. And there is no documentation ether.

The idea with the code = sha(source string) is very good, and would solve some problems from the past (too long string, cache busting etc.).

As you clear this, my assumptions and observations are wrong, but to be honest, I used the format x.y.z for all my translations strings so I never hit that problem.

Do you know, that you can write translation yaml files in the theme that you can then parse? We use this feature whenever we use message translation.

I would make a PR to the exporter/importer to fix this behavior.

matteotrubini:
> As you clear this, my assumptions and observations are wrong, but to be honest, I used the format x.y.z for all my translations strings so I never hit that problem.

I hit the problem every time as I need to include long contents. You can propose me to use content files (https://github.com/rainlab/translate-plugin/#content-translation), that are useful in some circumstances, but this method creates a lot of files that should be managed and consistency checks over time are more complicated.

> Do you know, that you can write translation yaml files in the theme that you can then parse? We use this feature whenever we use message translation.

Correct, but YAML it's not easily readable by my colleagues that perform content maintenance, and a even small change of source doesn't automatically invalidate the translation. And I expect this: anytime I can reinsert the same translation because it's still valid, but someone must had manage it.

Everytime, the most readable interface "for dummies" is a spreadsheet or a table with 2 columns: source | translation.

> I would make a PR to the exporter/importer to fix this behavior.

It would be fine to implement different file format as CSV has many limit. A good start would be support ODS - XLS(x) - JSON - Xliff (to be used with CAT tools). I think that to find well recognized parser/writer for those format would not be a problem.

munxar:
@matteotrubini No, I don't mean message translation, I mean a translation file:

*theme.yaml*
```yaml
name: 'Theme'
description: 'Theme for website'
author: Me
homepage: 'https://google.com'
code: my.theme
translate: config/lang.yaml
```

then write a translation file like this:
*config/lang.yaml*
```yaml
de:
    page.title: 'Meine Homepage'
en:
   page.title: 'My Homepage'
```

If you now hit 'Scan for Messages', the definitions from lang.yaml are parsed and imported.

munxar:
@matteotrubini look, I'm totally on your side, but can we do one thing after another?
So here I'd suggest we try to first patch your proposal  with the code, but your PR is not complete (not test) and my importer logic is wrong, after your imports about the code <-> default language.
If you don't mind, I'd make a new PR tomorrow that exposes the code.

after that I will try to prototype some sha/md5 stuff for the code, to prevent the discussed problems.

after that we should implement more export formats, like you proposed. Would this be a acceptable timeline for you?

LukeTowers:
@matteotrubini I would prefer discussions about code stay on GitHub, slack messages disappear after a short amount of time due to the 10,000 message limit, unless the conversation is just short back and forth and then a summary can be posted here.

munxar:
@LukeTowers @matteotrubini here you go:
https://github.com/rainlab/translate-plugin/pull/431

matteotrubini:
@LukeTowers Yes, in my vision a Slack group can be used to make brainstorming, then approved ideas should be summarized on GitHub issues, leaving coding related efforts here. The use of a Trello board to manage the project is optional.

LukeTowers:
@matteotrubini we'll stay way from trello boards for now. I've been experimenting with GitHub Projects (kanban organization of issues / PRs) for a bit now on the main repo and they seem fine.

@munxar can you check slack or send me an email? Have some questions for you

mjauvin:

Issue #418, opened 2018-11-01
TranslatableModel scopeTransWhere operator bug


aly-eldegwy:
In the TranslatableModel, the scopeTransWhere now supports the $operator, but in line 42, the $operator missed to be sent to the where() function of the table model. So, the result of the search is not the same when changing the language. As a suggestion, the fix may be by sending the $operator to the where like that :
     $q->where($this->model->getTable().'.'.$index, $operator, $value); // line 42

LukeTowers:
@aly-eldegwy please provide a PR with your proposed fix along with a detailed replication of the original issue along with how the original issue is fixed by your proposed fix.

mjauvin:
resolved by #433

PR #433, opened 2018-12-27
Add operator to where clause

Rike-cz:
Fix issue #418.

### DIFF
--- a/behaviors/TranslatableModel.php
+++ b/behaviors/TranslatableModel.php
@@ -39,7 +39,7 @@ class TranslatableModel extends TranslatableBehavior
         $query->select($this->model->getTable().'.*');
 
         $query->where(function($q) use ($index, $value, $operator) {
-            $q->where($this->model->getTable().'.'.$index, $value);
+            $q->where($this->model->getTable().'.'.$index, $operator, $value);
             $q->orWhere(function($q) use ($index, $value, $operator) {
                 $q
                     ->where('rainlab_translate_indexes.item', $index)


LukeTowers:
@munxar @osmanzeki could you take a look?

mjauvin:
@Rike-cz has this been tested so far? If you could add some Unit tests to help verify this, I would definitely merge this.

LukeTowers:
@mjauvin I'm not sure this could necessarily be unit tested. If @Rike-cz could provide an example use case that is broken in the current behaviour but fixed by this PR then that should be all that we need to get this merged (after testing it of course)

w20k:
Would be nice to have an example use case from @Rike-cz, and it's good to go.

Rike-cz:
That was only missing `$operator` in where clause when comparing with value as in line bellow, where input value is compared with value in translate table. I do not have use case for this now, but you can simply use `!=` or `LIKE` and compare it with some wildcards, for example: `$query->transWhere('slug', 'test', null, '!=')`.

mjauvin:
@rike-cz can you provide behavior before and after the change? Have you used the transWhere() method with this modification?

mjauvin:
Ok, so basically, before this change, since the $operator was NOT provided to the where() method of the query, the DEFAULT operator (=) for where() method was used.

With this change, as long as NO operator is provided to transWhere() method, the behavior will stay the same since the default value of the $operator argument to transWhere() is also "=".

The only difference is that we can now provide a different operator to transWhere and it will be passed to the where() method of the query.

LukeTowers:
@mjauvin please make sure you update the labels as appropriate after taking action on an issue / PR.

PR #436, opened 2018-12-28
fix translatable behavior event hooks

mjauvin:
The TranslatableBehavior class prevents the getter/setter for "translatable" dynamic property to work when hooking to the model.beforeGetAttribute / model.beforeSetAttribute events.

This PR fix the issue by ignoring the hook when the key == 'translatable"

### DIFF
--- a/classes/TranslatableBehavior.php
+++ b/classes/TranslatableBehavior.php
@@ -59,13 +59,13 @@ abstract class TranslatableBehavior extends ExtensionBase
         $this->initTranslatableContext();
 
         $this->model->bindEvent('model.beforeGetAttribute', function($key) {
-            if ($key != 'translatable' && $this->isTranslatable($key)) {
+            if ($key !== 'translatable' && $this->isTranslatable($key)) {
                 return $this->getAttributeTranslated($key);
             }
         });
 
         $this->model->bindEvent('model.beforeSetAttribute', function($key, $value) {
-            if ($key != 'translatable' && $this->isTranslatable($key)) {
+            if ($key !== 'translatable' && $this->isTranslatable($key)) {
                 return $this->setAttributeTranslated($key, $value);
             }
         });
--- a/classes/TranslatableBehavior.php
+++ b/classes/TranslatableBehavior.php
@@ -59,13 +59,13 @@ abstract class TranslatableBehavior extends ExtensionBase
         $this->initTranslatableContext();
 
         $this->model->bindEvent('model.beforeGetAttribute', function($key) {
-            if ($this->isTranslatable($key)) {
+            if ($key != 'translatable' && $this->isTranslatable($key)) {
                 return $this->getAttributeTranslated($key);
             }
         });
 
         $this->model->bindEvent('model.beforeSetAttribute', function($key, $value) {
-            if ($this->isTranslatable($key)) {
+            if ($key != 'translatable' && $this->isTranslatable($key)) {
                 return $this->setAttributeTranslated($key, $value);
             }
         });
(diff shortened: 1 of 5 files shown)


classes/TranslatableBehavior.php
```diff
@@ -59,13 +59,13 @@ public function __construct($model)
         $this->initTranslatableContext();
 
         $this->model->bindEvent('model.beforeGetAttribute', function($key) {
-            if ($this->isTranslatable($key)) {
+            if ($key != 'translatable' && $this->isTranslatable($key)) {
```
LukeTowers: Strict type checking please `!==` in both spots

LukeTowers:
@mjauvin was there an issue for this or was it just a comment in the Slack? Could you provide background for the fix and then @munxar could test it :)

w20k:
@LukeTowers, it was the issue from slack 😄

>From @amdad: On latest build 445 and latest dev rainlab Translate on lang switch I got a problem like this:
```Undefined property: Cms\Classes\Page::$translatable
/public_html/plugins/rainlab/translate/behaviors/TranslatablePage.php line 36

TYPE
Undefined

EXCEPTION
ErrorException
 
    }
  
    public function getTranslatableAttributes()
    {
        $attributes = [];
  
        foreach ($this->model->translatable as $attr) {
            $attributes[] = "settings[{$attr}]";
        }
        return $attributes;
    }
  
    public function getModelAttributes()```

**Comment by mjauvin** [comment]
image

**Comment by mjauvin** [comment]
Because of the model.beforeGetAttribute hook, the basic behavior of the Halcyon model was bypassed (Model::getAttribute returned the beforeGetAttribute hook's value instead of returning the value from the attributes array). So the dynamically added "translatable" attribute could not be called directly as in model->translatable but needed to be accessed using the attributes array (model->attributes['translatable']).

**Comment by LukeTowers** [comment]
@mjauvin could you bump the version.yaml for marketplace users?

File: updates/version.yaml [hunk 2]
```diff
@@ -52,3 +52,4 @@
 1.4.1: Updated Hungarian translation. Added Arabic translation. Fixed issue where default texts are overwritten by import. Fixed issue where the language switcher for repeater fields would overlap with the first repeater row.
 1.4.2: Add multilingual MediaFinder
 1.4.3: !!! Please update OctoberCMS to Build 444 before updating this plugin. Added ability to translate CMS Pages fields (e.g. title, description, meta-title, meta-description)
+1.4.4: Fix for the translatable dynamic property to be accessible in front-end.
```
LukeTowers: Change to `Fixed issue when using the language switcher on v1.4.4 of the plugin` please

munxar:
a unit test that covers the bug would be amazing!

mjauvin:
@munxar not sure how to test this.

munxar:
@mjauvin like this:
https://github.com/rainlab/translate-plugin/blob/master/tests/unit/behaviors/TranslatableCmsObjectTest.php

I'm not up to date which object you need to attach the translatable behavior, but this should be a good start.

Eoler:
_Meanwhile, back in the jungle..._ sites using static pages with 1.4.3 are broken.

LukeTowers:
@abdulla-allaith @Eoler can you test this PR and confirm that it works? I'll merge it once we have confirmation.

marcomessa:
Tested now on a fresh October installation and works flawlessy.

Eoler:
Tested on marketplace-installed edge build 466 (in production) - no more `undefined $translatable property` exception.

marcomessa:
But array_keys() expects parameter 1 to be array, object given is still present in staticpages plugin

LukeTowers:
@marcomessa does this PR fix the issue? Are you talking about a separate issue? Please provide more detail.

marcomessa:
I apologize, I arrived in this PR from [here](https://github.com/rainlab/translate-plugin/pull/367#issuecomment-451887154) and with another issue: after last update, I can't switch languages in my sites. This PR fixes this issue, but I noticed that the one discussed here (array_keys() expects parameter 1 to be array, object given) is still present when in bakcend I try to go to my staticpage plugin. All of this is happening in a fresh local install.

Balex70:
Hi, fix working great but I can't correctly update package with composer. Can someone update package on packageist.org (https://packagist.org/packages/rainlab/translate-plugin#dev-master). Looks like the last commit is outdated, f8a352335c0126a469a01f14c8d7c4f1f8d8aba6 instead of 16b82c5b1eaec53560981bb7b5ba8c1cbfb9ac50

LukeTowers:
@Balex70 done

Issue #439, opened 2019-01-07
pruneTranslatedContentTemplates: array_keys() expects parameter 1 to be array, object given


adrenth:
Caused by #432 

Calling array_keys on a Collection will never work in PHP 7.1.

```
$locales = LocaleModel::listAvailable();

$extensions = array_map(function($ext) {
    return '.'.$
}, array_keys($locales));
```

`$locales` contains a Collection (which is not an array).

Issue #438, opened 2019-01-07
error array_keys() expects parameter 1 to be array, object given in EventRegistry.php after edge update


Verbat:
I updated OctoberCMS and the translate Plugin (edge update enabled) to their latest version.
I had to apply the following fix https://github.com/rainlab/translate-plugin/pull/436 for the CMS pages which works fine.
I also noticed that when I try to open the Pages menu from the static Page Plugin in the backend I'll receive the following error on the line which says array_keys($locales));. 

array_keys() expects parameter 1 to be array, object given
C:\laragon\www\connect-com\plugins\rainlab\translate\classes\EventRegistry.php line 219

```
public function pruneTranslatedContentTemplates($templates)
    {
        $locales = LocaleModel::listAvailable();
        print_r($locales);
        $extensions = array_map(function($ext) {
            return '.'.$ext;
        }, array_keys($locales));
  
        return $templates->filter(function($template) use ($extensions) {
            return !Str::endsWith($template->getBaseFileName(), $extensions);
        });
    }
```
Screenshot below:
https://screenshotscdn.firefoxusercontent.com/images/1ae4663a-fc60-4783-9dca-00594ef6df82.png

adrenth:
See #439

LukeTowers:
Duplicate of #439.

PR #440, opened 2019-01-07
Fix replacement of `lists` to `pluck` -> array should be returned.

adrenth:
See #439 #438


LukeTowers:
@adrenth please also bump the version.yaml file for the marketplace users.

adrenth:
Done.

PR #453, opened 2019-03-12
Fix for #209

mjauvin:
When the model.saveInternal event is fired on a related model update without a form controller, translatable attributes are not set.

This calls the setAttributeTranslated() method for each entry in post('RLTranslate')

### DIFF
--- a/classes/TranslatableBehavior.php
+++ b/classes/TranslatableBehavior.php
@@ -240,9 +240,10 @@ abstract class TranslatableBehavior extends ExtensionBase
     {
         // empty translatableAttributes when called from a related model 
         if (!$this->translatableAttributes && post('RLTranslate')) {
+            $translatable = $this->model->getTranslatableAttributes();
             foreach (post("RLTranslate") as $locale => $data) {
                 foreach ($data as $key => $value) {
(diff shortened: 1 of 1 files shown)


alxy:
The real issue is that $translatableAttributes is empty in `syncTranslatableAttributes()`.
I have started to debug this case and it appears that the model instance is somehow not beeing reused by the FormController behavior, but a new model isntance is populated with the data. This process however does not take the behaviors into account.
image
 As you can see, these two are different instances (by comparing their IDs). The first one is dumped in `setAttributeTranslated` , the second one in `syncTranslatableAttributes`.

alxy:
Okay, I have an alternative fix, but it needs heavy (!) testing as it affects Octobers core.

Replace this line: https://github.com/octobercms/october/blob/master/modules/backend/behaviors/RelationController.php#L1065 with  ``$model = $this->manageWidget->model;`` and it should work. I have no idea if it has any other consequences.

mjauvin:
Nice work @alxy !

Can you create a new PR on core?

alxy:
Already done: https://github.com/octobercms/october/pull/4193

PR #454, opened 2019-03-20
Use translate.localePicker.translateParams in AlternateHrefLangElemets component

patrick-durrer:


components/AlternateHrefLangElements.php
```diff
@@ -3,6 +3,7 @@
 namespace Rainlab\Translate\Components;
 
 use Cms\Classes\ComponentBase;
+use Illuminate\Support\Facades\Event;
```
LukeTowers: Why do you fully specify that facade instead of just using `use Event`?

alxy:
I think, if this is accepted, the event needs a more generic name, as the `localePicker` part of it doesnt make too much sense in this case. Maybe `translate.translateParams` is already okay.

LukeTowers:
@alxy the event is already in use isn't it? Probably not going to change it now.

@patrick-durrer what is the purpose of this?

mjauvin:
@LukeTowers @bennothommo this PR would potentially solve the Rainlab.Blog slug translation issues in AlternateHrefLangElements component with help from https://github.com/rainlab/blog-plugin/pull/436

mjauvin:
I have tested this along with rainlab/blog-plugin#436 and I get the correct translated URLs in the added`<link rel="alternate" hreflang=...`  in <head> section of the page.

I would be inclined to merge this.

mjauvin:
@patrick-durrer @tobias-kuendig could you apply @LukeTowers' requested fix to the facade?

PR #461, opened 2019-03-30
prevent nested field names from matching Model->translatable[] fields

### LINKED ISSUE #459: Repeater fields are wrongly displayed as translatable
I have the following model:

```php
class Product extends Model
{
    public $implement = ['@RainLab.Translate.Behaviors.TranslatableModel'];
    public $translatable = [
        'name',
    ];
}
```

with the following form:

```yaml
fields:
    name:
        label: Name
        type: text

additional_properties:
    label: Additional properties
    type: repeater
    form:
        fields:
            name:
                label: Name
                type: text
            value:
                label: Value
                type: text
```

Since the repeater field is called `name` and `name` is marked as `translatable` I end up with this:

image

This isn't what I was expecting, since the repeater field is actually "called" `additional_properties.name`.

What makes matters worse is that when I hit save, the model's `name` value gets updated to what was entered in the repeaters `name` field (in this case `"Property name"`).

To me this is unexpected behaviour. The translatable fields should be aware of their context (being inside a repeater). What do you think?

mjauvin:
Repeater and NestedForm formwidgets' field names are marked as translatable when they should when their name matches another field name from the model.

Fix #459

### DIFF
--- a/classes/EventRegistry.php
+++ b/classes/EventRegistry.php
@@ -87,7 +87,7 @@ class EventRegistry
             return;
         }
 
-        if (!empty($widget->config->fields) && !$widget->config->isNested) {
+        if (!empty($widget->config->fields) && !$widget->isNested) {
             $widget->fields = $this->processFormMLFields($widget->fields, $model);
         }
 
--- a/classes/EventRegistry.php
+++ b/classes/EventRegistry.php
@@ -87,7 +87,7 @@ class EventRegistry
             return;
         }
 
-        if (!empty($widget->config->fields)) {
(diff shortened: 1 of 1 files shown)


mjauvin:
We should merge this as well, as soon as possible.

tobias-kuendig:
Just tested this with the test plugin meantioned in https://github.com/rainlab/translate-plugin/issues/459 on build 451. This PR resolved the issue for me.

mjauvin:
@LukeTowers can we have this merged?

PR #463, opened 2019-04-01
Fix post value rewrite

acasar:
Due to a recent change in https://github.com/octobercms/library, changing the $_POST superglobal directly doesn't work any more. 

The issue is described more in detail here: https://github.com/octobercms/library/commit/fd839047c2d7757f225a8ed60673ca8a20fe7df4


mjauvin:
Nice work @acasar 

Please change ALL occurences of the $_POST array (Line 69)

w20k:
@acasar could you fix all occurrences where $_POST/$_GET used ?) Thanks!

acasar:
Ah, didn't even see those :) I think I got them all now.

daftspunk:
Great work, thanks for this!

PR #466, opened 2019-04-06
Bring MLRepeater inline with core changes to Repeater

LukeTowers:
Related: https://github.com/octobercms/october/pull/4234

### DIFF
--- a/formwidgets/MLRepeater.php
+++ b/formwidgets/MLRepeater.php
@@ -117,7 +117,7 @@ class MLRepeater extends Repeater
 
         $this->formField->value = $lockerData;
 
-        $this->reprocessExistingLocaleItems($lockerData);
+        $this->reprocessLocaleItems($lockerData);
 
         foreach ($this->formWidgets as $key => $widget) {
             $value = array_shift($lockerData);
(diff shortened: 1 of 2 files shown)


bennothommo:
@LukeTowers I've tested the changes both in here and https://github.com/octobercms/october/pull/4234 with the Test Plugin (and the additional test cases in https://github.com/octoberrain/test-plugin/pull/56 for MLRepeater usage).

I noticed a few things whilst testing the **Content** repeater of the Countries form in the Test Plugin:

- It appears that any translated data is not saving - only the default language's data is saved.
- If you use the language selector above the repeater to toggle between each language available, the data is lost between the change.
- MLRepeater widget items don't seem to be able to be deleted. Refreshing the page brings the deleted items back.

LukeTowers:
@w20k can you test this too? @BennoThommo would you be willing to look into those issues and see if you can find the fixes for them and make a PR to the branch?

w20k:
@LukeTowers sure, will run few tests!

bennothommo:
@LukeTowers sure thing

bennothommo:
@LukeTowers I've started a couple of branches to fix the issues I reported above:

https://github.com/octobercms/october/pull/4242
https://github.com/rainlab/translate-plugin/pull/467

I've sorted the first two issues in that list, and the third appears to have resolved itself. However, I'm looking at another issue where re-ordering the items in an MLRepeater seems to not apply across language changes. Hopefully will have that sorted within the next day or so.

CptMeatball:
@bennothommo Just a quick question about the reordering not applying across languages. That sounds like a feature not a bug, as the content of each locale is unrelated to the other locales. In that way reordering **should not** apply on language switch. 

Or you mean that switching languages without saving causes the rearrange to be lost. But just wanted to get this out there.

bennothommo:
@CptMeatball As far as I can tell, the MLRepeater widget seems to operate under the assumption that whatever items are in the repeater are used for all languages, and that it's simply the content of those items that changes between languages.

CptMeatball:
Yeah that's how I use it too, but over at Slack I heard people say that , and I quote:

>I have many situations where the content structure is different for locales. It depends on customer requirements and many customers prefer one language as main and the other as complementary...

That's why I'm in favor of a sync option between the locales, but that's another PR all together.

bennothommo:
I'll probably need @LukeTowers or one of the core team to let us know on what is the intended operation of the MLRepeater widget. If it's the case that the locales are allowed different items, then it won't need my re-order fix that's in progress.

LukeTowers:
@bennothommo @CptMeatball @w20k content should be able to be different, so the reordering shouldn't be needed. It shouldn't be having empty items when switching to a new language though either.

CptMeatball:
I noticed that recently. That the structure is copied to the secondary languages. Wasn't that different before?

LukeTowers:
Probably @CptMeatball, I'm not sure what caused that, but we need to stop that behaviour.

w20k:
That was a bug for a title, that's for sure!

LukeTowers:
@CptMeatball @munxar @mjauvin could you test @bennothommo's latest changes in both PRs?

mariavilaro:
I tested it and it works ok for me. But I'm one of those that would like to be able to have the same elements in all locales, just with the texts translated. In my case, I use the repeater in a portfolio, for the credits, something like:

    Title - Person
    ---------------
    Music - Someone
    Choreography - Someone else

So it would be perfect to be able to write everything in the default language and then just translate the needed parts (in this case only the titles because the people names are the same)

PR #469, opened 2019-04-16
MLFileUpload widget enables translation of file captions

mariavilaro:
I added a widget that enables translation of attached files captions (title and description). I used this code https://github.com/Glitchbone/octobercms-filetranslate-plugin with some changes and fixes.


mjauvin:
Nice but there is already such a PR:

https://github.com/rainlab/translate-plugin/pull/442

mariavilaro:
Yes I just found it 🤦‍♀️

PR #472, opened 2019-04-18
implement PurgeableBehavior

mjauvin:


Plugin.php
```diff
@@ -54,6 +54,7 @@ public function register()
          */
         File::extend(function ($model) {
             $model->addDynamicProperty('translatable', ['title', 'description']);
+            $model->extendClassWith('October\Rain\Database\Behaviors\Purgeable');
             $model->extendClassWith('RainLab.Translate.Behaviors.TranslatableModel');
```
LukeTowers: Do we need to add a `purgeable` property? Also, could you make the style of both these imports match (`\` vs `.`)? Don't really care which one, just pick one

mjauvin: no, the purgeable property gets added by the Behavior and it automatically add the dynamicProperties to the purgeable property

LukeTowers: Sweet

mjauvin:
Luke, can these be merged in the wip branch so that it's easier for people to test it?

PR #478, opened 2019-05-03
Make sluggable work with translatable slugs

mariavilaro:
Generate slugs for models that implement the Sluggable Trait and have translatable slugs

### DIFF
--- a/traits/Sluggable.php
+++ b/traits/Sluggable.php
@@ -1,6 +1,6 @@
 <?php namespace RainLab\Translate\Traits;
 
-use October\Rain\Support\Str;
+use Str;
 
 trait Sluggable
 {
--- a/traits/Sluggable.php
+++ b/traits/Sluggable.php
@@ -10,7 +10,9 @@ trait Sluggable
      */
     protected function slugAttributesTranslated($locale)
     {
-        if (!isset($this->model->slugs) || !method_exists($this->model, 'slugAttributes')) return;
+        if (!isset($this->model->slugs) || !method_exists($this->model, 'slugAttributes')) {
+            return;
+        }
         
         $optionedAttributes = $this->getTranslatableAttributesWithOptions();
 
--- a/traits/Sluggable.php
+++ b/traits/Sluggable.php
@@ -17,7 +17,9 @@ trait Sluggable
         $optionedAttributes = $this->getTranslatableAttributesWithOptions();
 
         foreach ($this->model->slugs as $slugAttribute => $sourceAttributes) {
-            if (!isset($optionedAttributes[$slugAttribute])) continue;
+            if (!isset($optionedAttributes[$slugAttribute])) {
+                continue;
+            }
             $this->setSluggedValueTranslated($locale, $slugAttribute, $sourceAttributes);
         }
     }
--- a/traits/Sluggable.php
+++ b/traits/Sluggable.php
@@ -1,0 +1,113 @@
+<?php namespace RainLab\Translate\Traits;
+
+use October\Rain\Support\Str;
+
+trait Sluggable
+{
+    /**
+     * Adds translatable slug attributes to the translation data
+     * @return void
+     */
+    protected function slugAttributesTranslated($locale)
+    {
+        if (!isset($this->model->slugs) || !method_exists($this->model, 'slugAttributes')) return;
+        
+        $optionedAttributes = $this->getTranslatableAttributesWithOptions();
+
+        foreach ($this->model->slugs as $slugAttribute => $sourceAttributes) {
+            if (!isset($optionedAttributes[$slugAttribute])) continue;
+            $this->setSluggedValueTranslated($locale, $slugAttribute, $sourceAttributes);
+        }
+    }
+
+    /**
+     * Sets a single translatable slug attribute value.
+     * @param string $locale
+     * @param string $slugAttribute Attribute to populate with the slug.
+     * @param mixed $sourceAttributes Attribute(s) to generate the slug from.
+     * Supports dotted notation for relations.
+     * @param int $maxLength Maximum length for the slug not including the counter.
+     * @return string The generated value.
+     */
+    protected function setSluggedValueTranslated($locale, $slugAttribute, $sourceAttributes, $maxLength = 175)
+    {
+        if (!isset($this->translatableAttributes[$locale][$slugAttribute]) || !mb_strlen($this->translatableAttributes[$locale][$slugAttribute])) {
+            if (!is_array($sourceAttributes)) {
+                $sourceAttributes = [$sourceAttributes];
+            }
+
+            $slugArr = [];
+            foreach ($sourceAttributes as $attribute) {
+                $slugArr[] = $this->getSluggableSourceAttributeValueTranslated($locale, $attribute);
+            }
+
+            $slug = implode(' ', $slugArr);
+            $slug = mb_substr($slug, 0, $maxLength);
+            $slug = Str::slug($slug, $this->model->getSluggableSeparator());
+        }
+        else {
+            $slug = $this->translatableAttributes[$locale][$slugAttribute];
+        }
+
+        return $this->translatableAttributes[$locale][$slugAttribute] = $this->getSluggableUniqueAttributeValueTranslated($locale, $slugAttribute, $slug);
+    }
+
+    /**
+     * Ensures a unique attribute value for the locale, if the value is already used a counter suffix is added.
+     * @param string $locale
+     * @param string $name The database column name.
+     * @param value $value The desired column value.
+     * @return string A safe value that is unique.
+     */
+    protected function getSluggableUniqueAttributeValueTranslated($locale, $name, $value)
+    {
+        $counter = 1;
+        $separator = $this->model->getSluggableSeparator();
+        $_value = $value;
+
+        while (($this->model->methodExists('withTrashed') && $this->model->allowTrashedSlugs) ?
+            $this->newSluggableQueryTranslated()->transWhere($name, $_value, $locale)->withTrashed()->count() > 0 :
+            $this->newSluggableQueryTranslated()->transWhere($name, $_value, $locale)->count() > 0
+        ) {
+            $counter++;
+            $_value = $value . $separator . $counter;
+        }
+
+        return $_value;
+    }
+
+    /**
+     * Returns a query that excludes the current record if it exists
+     * @return Builder
+     */
+    protected function newSluggableQueryTranslated()
+    {
+        return $this->model->exists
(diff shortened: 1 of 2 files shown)


mjauvin:
Can you describe the exact context where this can be used?

mariavilaro:
If you have a model that implements the Sluggable Trait, and with a translatable slug:

    class User extends Model
    {
        use \October\Rain\Database\Traits\Sluggable;

        protected $slugs = ['slug' => 'name'];

        public $translatable = [ 'name', ['slug', 'index' => true] ];
}

When you save the model, only the slug for the default language is generated

mjauvin:
Is there any issues this PR solves?

@w20k @bennothommo @osmanzeki can you test this and let me know if that works as expected?

bennothommo:
@mariavilaro It might be better to make the new Sluggable trait extend the core `October\Rain\Database\Traits\Sluggable` class and adjust it to handle the translated slug properties as necessary, as there seems to be a lot of repeated code here.

It also should be an optional trait that can be used on translatable models, and not forced by the TranslatableModel behavior.

traits/Sluggable.php
```diff
@@ -0,0 +1,113 @@
+<?php namespace RainLab\Translate\Traits;
+
+use October\Rain\Support\Str;
```
bennothommo: @mariavilaro This can be shortened to just use the facade :)
```suggestion
use Str;
```

traits/Sluggable.php
```diff
@@ -0,0 +1,113 @@
...
+
+trait Sluggable
+{
+    /**
+     * Adds translatable slug attributes to the translation data
+     * @return void
+     */
+    protected function slugAttributesTranslated($locale)
+    {
+        if (!isset($this->model->slugs) || !method_exists($this->model, 'slugAttributes')) return;
```
bennothommo: @mariavilaro Please add braces around any condition blocks, as per [PSR-2](https://www.php-fig.org/psr/psr-2/) guidelines.
```suggestion
        if (!isset($this->model->slugs) || !method_exists($this->model, 'slugAttributes')) {
            return;
        }
```

traits/Sluggable.php
```diff
@@ -0,0 +1,113 @@
...
+        
+        $optionedAttributes = $this->getTranslatableAttributesWithOptions();
+
+        foreach ($this->model->slugs as $slugAttribute => $sourceAttributes) {
+            if (!isset($optionedAttributes[$slugAttribute])) continue;
```
bennothommo: @mariavilaro Please add braces around any condition blocks, as per PSR-2 guidelines.

mariavilaro:
> @mariavilaro It might be better to make the new Sluggable trait extend the core `October\Rain\Database\Traits\Sluggable` class and adjust it to handle the translated slug properties as necessary, as there seems to be a lot of repeated code here.

The core Sluggable class is already extended by the model, and their methods are already used by the default language slug generation, and the new trait also use some of them (e.j. $this->model->getSluggableSeparator()). I can't just override the methods, or then the core Sluggable class fails when the default language slug is generated. If there is a simpler way of doing this, I don't know how to do it.

> It also should be an optional trait that can be used on translatable models, and not forced by the TranslatableModel behavior.

If you want to use the Sluggable behavior, and also to have translated slugs, you're saying that you want your slugs to be generated, IMHO. The problem now is that the alternate language slugs are not automatically generated, and if you write them manually, they are not checked for uniqueness. Anyway, if you think that somebody could want to have translatable slugs that are not automatically populated or checked, maybe we could indicate that we want our slugs to behave as slugs by doing something like this:

    public $translatable = [ 'name', ['slug', 'index' => true, 'slug' => true] ];

But again, I think this is redundant.

mariavilaro:
> Is there any issues this PR solves?

The problem now is that the alternate language slugs are not automatically generated, and if you write them manually, they are not checked for uniqueness

bennothommo:
@mariavilaro Thanks for your response. There's a couple of things I'm suggesting we should avoid here:

- There's a reliance on the `October\Rain\Database\Traits\Sluggable` trait in your new Sluggable trait without it being explicitly defined as a dependency. If a developer who uses this new trait forgets to include the original trait, they may be confused as to why it is not working. This could be resolved by documentation  of course, but it would be cleaner if the new trait explicitly included the original trait in some way (eg. add a `use` declaration for the original Sluggable trait in the new Sluggable trait) and just use the new trait in any models that need translatable slug support.
- Given that both this new Sluggable trait and the old Sluggable trait share a lot of the same code but are completely independent, any future changes to the Sluggable trait are now possibly going to have to be made in two places, increasing the maintenance burden.
- Having it defined in the TranslatableModel behavior by default effectively means that additional functionality is being inserted into any model using the TranslatableModel behavior, even if they are not intending to use slugs. It would be better if this only occurs when needed.

Happy to hear your thoughts on the above.

LukeTowers:
I agree with @bennothommo's comments

LukeTowers:
To be clear, I would like to see an implementation closer to https://github.com/rainlab/translate-plugin/pull/478#issuecomment-489316751 before merging this.

mariavilaro:
So @bennothommo you suggest to make a full working Sluggable trait, which extends Sluggable and which takes care of all slugs, both the default language and alternate languages.

So the model would use it like this:

    class User extends Model
    {
        use \RainLab\Translate\Traits\Sluggable;
        protected $slugs = ['slug' => 'name'];
        public $implement = ['RainLab.Translate.Behaviors.TranslatableModel'];
        public $translatable = [ 'name', ['slug', 'index' => true] ];
    }

Do I understand it right?

bennothommo:
@mariavilaro That's correct. The new `RainLab\Translate\Traits\Sluggable` trait should be an extension of the functionality that already occurs with the original `October\Rain\Database\Traits\Sluggable` trait in which it handles the translated slugs, but also allows the original functionality to handle the default language.

The developer should only need to include this new trait if they want both the default and translated language slugs saved. If they want the original functionality where it only saves the default lanugage, they can continue to use the original trait.

mjauvin:
@mariavilaro any progress on this?

mjauvin:
@mariavilaro are you still interested in pursuing this?

mjauvin:

mariavilaro:
Hi sorry, I tried to make the independent RainLab\Translate\Traits\Sluggable and didn't succeed, I needed it to work urgently so I just made it work in my plugin. I don't really know how to make it work as a Translate Trait :(

mariavilaro:
If someone wants to try it, this is the last try I did

```php
<?php namespace RainLab\Translate\Behaviors;

use System\Classes\ModelBehavior;
use Exception;
use Str;

class Sluggable extends ModelBehavior
{
    protected $translatedSlugsChanged = false;

    public function __construct($model)
    {
        $this->model = $model;

        $this->model->bindEvent('model.afterSave', function() {
            $this->afterSave();
        });
    }

    public function afterSave()
    {
        $locales = \RainLab\Translate\Models\Locale::listEnabled();
        $locales = array_keys($locales);
        $default_locale = \RainLab\Translate\Classes\Translator::instance()->getDefaultLocale();

        if (!method_exists($this->model, 'slugAttributes')) {
            throw new Exception(sprintf(
                'You must use the Sluggable trait in %s to use the Translate plugin Sluggable behaviour.', get_class($this->model)
            ));
        }

        if (!isset($this->model->slugs)) {
            throw new Exception(sprintf(
                'You must define a public $slugs property in %s to use the Translate plugin Sluggable behaviour.', get_class($this->model)
            ));
        }

        $optionedAttributes = $this->model->getTranslatableAttributesWithOptions();

        foreach ($this->model->slugs as $slugAttribute => $sourceAttributes) {
            if (!isset($optionedAttributes[$slugAttribute])) {
                throw new Exception(sprintf(
                    '$slugAttribute must be defined as an indexed translatable property in %s to use the Translate plugin Sluggable behaviour.', get_class($this->model)
                ));
            }
            foreach($locales as $locale) {
                if ($locale == $default_locale) {
                    continue;
                }
                $this->model->noFallbackLocale();
                $this->setSluggedValue($locale, $slugAttribute, $sourceAttributes);
            }
        }

        if ($this->translatedSlugsChanged) {
            //DOESN'T WORK
            //$this->model->save();
        }
    }

    protected function setSluggedValue($locale, $slugAttribute, $sourceAttributes, $maxLength = 175)
    {
        if (!$this->model->getAttributeTranslated($slugAttribute, $locale)) {
            if (!is_array($sourceAttributes)) {
                $sourceAttributes = [$sourceAttributes];
            }

            $slugArr = [];
            foreach ($sourceAttributes as $attribute) {
                $slugArr[] = $this->getSluggableSourceAttributeValue($locale, $attribute);
            }

            $slug = implode(' ', $slugArr);
            $slug = mb_substr($slug, 0, $maxLength);
            $slug = Str::slug($slug, $this->model->getSluggableSeparator());
        }
        else {
            $slug = $this->model->getAttributeTranslated($slugAttribute, $locale);
        }

        $unique_slug = $this->getSluggableUniqueAttributeValue($locale, $slugAttribute, $slug);

        if ($unique_slug != $this->model->getAttributeTranslated($slugAttribute, $locale)) {
            $this->translatedSlugsChanged = true;
            return $this->model->setAttributeTranslated($slugAttribute, $unique_slug, $locale);
        }

        return $unique_slug;
    }

    protected function getSluggableUniqueAttributeValue($locale, $name, $value)
    {
        $counter = 1;
        $separator = $this->model->getSluggableSeparator();
        $_value = $value;

        while (($this->model->methodExists('withTrashed') && $this->model->allowTrashedSlugs) ?
            $this->newSluggableQuery()->transWhere($name, $_value, $locale)->withTrashed()->count() > 0 :
            $this->newSluggableQuery()->transWhere($name, $_value, $locale)->count() > 0
        ) {
            $counter++;
            $_value = $value . $separator . $counter;
        }

        return $_value;
    }

    public function newSluggableQuery()
    {
        return $this->model->exists
            ? $this->model->newQuery()->where($this->model->getTable().'.'.$this->model->getKeyName(), '<>', $this->model->getKey())
            : $this->model->newQuery();
    }

    protected function getSluggableSourceAttributeValue($locale, $key)
    {
        if (strpos($key, '.') === false) {
            return $this->model->getAttributeTranslated($key, $locale);
        }

        $keyParts = explode('.', $key);
        $value = $this->model;
        foreach ($keyParts as $part) {
            if (!isset($value[$part])) {
                return null;
            }

            $value = $value[$part];
        }

        return $value;
    }
}
```

bennothommo:
@mariavilaro Given that it was me who laid down the initial challenge, I'll be happy to give it a go. Thanks a lot for your effort so far.

bennothommo:
Work continuing in #535.

Issue #346, opened 2018-02-05
Static Pages > Translated pages (files) not deleted when removing static page via the CMS


ranmichael:
**Build 431
Translate | 1.3.8 | 
Pages | 1.2.18 |** 

Steps to reproduce
- Create a static page (en + fr)
- Add some content and translated content in a second language
- Delete the page
Notice bug:
- The file in the primary language was deleted
- The file in the secondary language still exists

Also, sometimes I see traces of deleted files still exist under the meta/static-pages.yaml
This is a separate issue and need further investigating.

munxar:
I can reproduce this, non default language page is still present under content/static-pages-xy/abc.htm
I'll have a look into it.

LukeTowers:
Thanks @munxar!

PR #479, opened 2019-05-07
delete translation static content when parent page is deleted

mjauvin:
Fix for #346

### DIFF
--- a/behaviors/TranslatableCmsObject.php
+++ b/behaviors/TranslatableCmsObject.php
@@ -1,6 +1,5 @@
 <?php namespace RainLab\Translate\Behaviors;
 
-use File;
 use RainLab\Translate\Models\Locale;
 use RainLab\Translate\Classes\Translator;
 use RainLab\Translate\Classes\TranslatableBehavior;
--- a/behaviors/TranslatableCmsObject.php
+++ b/behaviors/TranslatableCmsObject.php
@@ -46,16 +46,15 @@ class TranslatableCmsObject extends TranslatableBehavior
             return $this->overrideTwigCacheKey($key);
         });
 
+        // delete all translation files associated with the default language static page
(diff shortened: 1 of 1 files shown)


mjauvin:
Please comment if you think there is a cleaner way to do this or if I am missing something.

behaviors/TranslatableCmsObject.php
```diff
@@ -43,6 +45,20 @@ public function __construct($model)
...
+            $contentPath = sprintf('%s/%s/%s', themes_path(), $model->theme->getDirName(), $model->getObjectTypeDirName());
+
+            foreach (Locale::listEnabled() as $locale => $label) {
+                if ($locale == $this->translatableDefault) {
+                    continue;
+                }
+                $localePath = sprintf('%s-%s/%s', $contentPath, $locale, $model->fileName);
+                if (File::exists($localePath)) {
+                    File::delete($localePath);
+                }
```
LukeTowers: Don't use the file facade for this, get the relevant Halcyon object directly and call delete on that; that will ensure support for the upcoming DB layer. Also add a one line comment describing the purpose of this code.

mjauvin: This should be the right way

LukeTowers: Did you test it?

LukeTowers: Why is it required to set the filename if you're calling `getCmsObjectForLocale` already?

mjauvin: Yes, tested it and it works fine.

As for setting the filename, I replicated code from another method ( syncTranslatableFileNames ) which does the same thing... but you are right, this is not necessary.

mjauvin:
@mariavilaro @osmanzeki @munxar can someone test this, it's working for me but I prefer having a second opinion.

PR #486, opened 2019-05-22
add option to disable locale prefix routes

### LINKED ISSUE #476: Option to disable prefixed routes
Option to disable routing in routes.php (https://github.com/rainlab/translate-plugin/blob/master/routes.php#L9) to prevent duplicate URLs for SEO if is used custom plugin to set language or language is set depending on browser.

mjauvin:
New config to disable locale prefix routes (disableLocalePrefixRoutes). Defaults to false.

Resolves #476

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -8,7 +8,6 @@ return [
     |
     | Should the default locale be prefixed by the plugin?
     |
-    |
     */
     'prefixDefaultLocale' => true,
 
@@ -30,7 +29,8 @@ return [
     | Disable locale prefix routes
     |--------------------------------------------------------------------------
     |
-    | Should we disable the locale prefix routes?
+    | Disables the automatically generated locale prefixed routes
+    | (i.e. /en/original-route) when enabled.
     |
     */
     'disableLocalePrefixRoutes' => false,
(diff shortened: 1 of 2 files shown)


config/config.php
```diff
@@ -26,4 +25,14 @@
     */
     'cacheTimeout'        => 1440,
 
+    /*
+    |--------------------------------------------------------------------------
+    | Disable locale prefix routes
+    |--------------------------------------------------------------------------
+    |
+    | Should we disable the locale prefix routes?
```
LukeTowers: Please change to 

> Disables the automatically generated locale prefixed routes (i.e. /en/original-route) when enabled.

config/config.php
```diff
@@ -8,7 +8,6 @@
     |
     | Should the default locale be prefixed by the plugin?
     |
-    | ( For now only used in the alternate-hreflang component )
     |
```
LukeTowers: Please remove another line

routes.php
```diff
@@ -22,6 +22,10 @@
         return;
     }
 
+    if (Config::get('rainlab.translate::disableLocalePrefixRoutes', false)) {
```
Samuell1: I think this can be moved on top of this function to prevent useless calls for App and Translator.

mjauvin:
@Samuell1 can you test this please?

Samuell1:
@mjauvin Tested and it works 👍

PR #482, opened 2019-05-13
Fixes issue with indexing

panakour:
Reference #481 

Peek 2019-05-13 09-21

october:454
translate: 1.6.0
pages: 1.2.21

PR #491, opened 2019-06-21
fix #482 issue

mjauvin:
make sure $value is an array when using array_values

### DIFF
--- a/formwidgets/MLRepeater.php
+++ b/formwidgets/MLRepeater.php
@@ -64,7 +64,7 @@ class MLRepeater extends Repeater
     {
         $this->rewritePostValues();
 
-        return $this->getLocaleSaveValue(array_values($value));
+        return $this->getLocaleSaveValue(is_array($value) ? array_values($value) : $value);
     }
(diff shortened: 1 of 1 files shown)


mjauvin:
@haralake try this fix

haralake:
Yes that fixes the error. When i switch language, create a field and save it data saved properly, but for some weird reason when i delete the field i created and save it data doesn't removed, are still here.

mjauvin:
Not sure I understand what you mean.

haralake:
Let me explain:
1 .On static page switch language with language selector
2. Create one field using repeater field and save page
3. Delete this field you have created and save page
4. On static page content that keeps fields data you will see that data are still there, they have not been deleted

mjauvin:
Ok, I didn't test this yet, but does it create problems other than having left-over data?

Would you consider this a seperate issue? If so, can you open a new issue for this?

haralake:
Sorry for delay! Ok i will create a new issue for this, thank you.

PR #496, opened 2019-07-04
Fix for rainlab/pages-plugin#404

LukeTowers:


mjauvin:
@LukeTowers why is this marked as work in progress?

LukeTowers:
@mjauvin it never got tested by the original issue submitter and I don't actually use this plugin so I haven't fully tested it. If you can verify the original issue and that this fix solves it without breaking anything feel free to merge it.

mjauvin:
@LukeTowers This PR is using Lang::getLocale() and App::setLocale() methods... shouldn't it use Rainlab\Translate\Models\Locale and Rainlab\Translate\Classes\Translator classes methods ?

LukeTowers:
@mjauvin it's been a long time since I last looked into this, but this gets triggered when editing content in the backend, which I believe the translate plugin interacts with by using the App locale and not those other classes. Feel free to look into it further though.

Issue #502, opened 2019-07-30
Eager loading attributes


hirisov:
Currently the plugin issues a new SQL query for each items in a collection to fetch the data from rainlab_translate_attributes table. 

For example if I have a collection with 3 items of H3Kft\Pcd\Models\Woa (with id's if 1, 2, 3) and want to get the "hu" translations this will result in 3 separate queries:

select * from `rainlab_translate_attributes` where `locale` = 'hu' and `model_id` = 1 and `model_type` = 'H3Kft\Pcd\Models\Woa' limit 1 

select * from `rainlab_translate_attributes` where `locale` = 'hu' and `model_id` = 2 and `model_type` = 'H3Kft\Pcd\Models\Woa' limit 1 

select * from `rainlab_translate_attributes` where `locale` = 'hu' and `model_id` = 3 and `model_type` = 'H3Kft\Pcd\Models\Woa' limit 1 

instead I think it would be ideal to execute one query:

select * from `rainlab_translate_attributes` where `locale` = 'hu' and `model_id` in (1, 2, 3) and `model_type` = 'H3Kft\Pcd\Models\Woa' 

This would result in a major performance boost for non default languages. 

I would be happy to offer bounty for such an improvement, i think others would join, as this is now a major bottleneck in performance of multilingual sites.

bennothommo:
@hirisov We would definitely appreciate any bounties you and others are willing to offer to get the feature implemented. Feel free to post a bounty here: https://www.bountysource.com/issues/77949687-eager-loading-attributes

hirisov:
Thanks for creating the bounty, i backed it up. I hope others will join! :)

buccia:
👍

hirisov:
A3Emanuele Bucciarelli <notifications@github.com> ezt írta (2019. aug. 1. 9:40):👍

bunker-condenast:
It would be a very appreciated improvement!

massimomegistus:
Hi guys, i've just developed this feature in a very simple way.
It seems to work well. Please test it and let me know
https://github.com/rainlab/translate-plugin/pull/504

mosid:
@hirisov @bennothommo  isn't the messages saved to cache?
Why bothering optimizing the query?

mjauvin:
@massimomegistus did you pickup the bounty on this?

massimomegistus:
@mjauvin not yet. I'm going to try :)

PR #504, opened 2019-08-05
Attribute model and translations morphMany relation

massimomegistus:
TranslatableBehavior with a "translations" morphMany relation to the new Attribute model, so you can use the Eloquent eager loading, for example:

```php
public function scopeList($query)
     {
         return $query->with([
           'translations',
           'featured_image',
           'featured_image.translations',
         ]);
     }
```

### DIFF
--- a/behaviors/TranslatableModel.php
+++ b/behaviors/TranslatableModel.php
@@ -31,15 +31,6 @@ class TranslatableModel extends TranslatableBehavior
         ];
     }
 
-    public function scopeTransWith($query)
-    {
-        $query->with([
-          'translations'
-        ]);
-
-        return $query;
-    }
-
     /**
      * Applies a translatable index to a basic query. This scope will join the index
      * table and cannot be executed more than once.
--- a/behaviors/TranslatableModel.php
+++ b/behaviors/TranslatableModel.php
@@ -227,11 +227,11 @@ class TranslatableModel extends TranslatableBehavior
             return $this->translatableAttributes[$locale] = [];
         }
 
-        $obj = $this->model->translations->first(function ($value, $key) use($locale) {
+        $obj = $this->model->translations->first(function ($value, $key) use ($locale) {
             return $value->attributes['locale'] === $locale;
         });
 
-        $result = $obj ? json_decode($obj['attribute_data'], true) : [];
+        $result = $obj ? json_decode($obj->attribute_data, true) : [];
 
         return $this->translatableOriginals[$locale] = $this->translatableAttributes[$locale] = $result;
     }
--- a/behaviors/TranslatableModel.php
+++ b/behaviors/TranslatableModel.php
@@ -21,6 +21,25 @@ use Exception;
  */
 class TranslatableModel extends TranslatableBehavior
 {
+    public function __construct($model)
+    {
+        parent::__construct($model);
+
+        $model->morphMany['translations'] = [
+            'RainLab\Translate\Models\Attribute',
+            'name' => 'model'
+        ];
+    }
+
+    public function scopeTransWith($query)
+    {
+        $query->with([
+          'translations'
+        ]);
+
+        return $query;
+    }
+
     /**
      * Applies a translatable index to a basic query. This scope will join the index
      * table and cannot be executed more than once.
@@ -84,7 +103,7 @@ class TranslatableModel extends TranslatableBehavior
             return;
         }
 
-        /** 
+        /**
          * @event model.translate.resolveComputedFields
          * Resolve computed fields before saving
          *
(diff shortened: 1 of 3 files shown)


w20k:
@massimomegistus could you also add unit tests to a `TranslatableModelTest.php`?

massimomegistus:
@w20k unfortunately I have an issue running phpunit from rainlab/translate folder . I get this error
Cannot open file "/var/www/public/octest/plugins/rainlab/translate/../../../tests/bootstrap.php"
Any idea? Thanks for the support.

w20k:
@massimomegistus the easiest way would be to add `translate-plugin` to a live|active project with everything setup. `Composer update` and run tests from the root directory, like so: `php ./vendor/bin/phpunit plugins/rainlab/translate/tests/unit`.

massimomegistus:
It works, thank you! I'll add the unit test very soon (by the end of the week)

w20k:
Cool, thanks! Once you're done, will revise it! ;)

behaviors/TranslatableModel.php
```diff
@@ -208,13 +227,11 @@ protected function loadTranslatableData($locale = null)
             return $this->translatableAttributes[$locale] = [];
         }
 
-        $obj = Db::table('rainlab_translate_attributes')
-            ->where('locale', $locale)
-            ->where('model_id', $this->model->getKey())
-            ->where('model_type', get_class($this->model))
-            ->first();
+        $obj = $this->model->translations->first(function ($value, $key) use($locale) {
```
bennothommo: @massimomegistus Please add a space between `use` and `(`:
```suggestion
        $obj = $this->model->translations->first(function ($value, $key) use ($locale) {
```

behaviors/TranslatableModel.php
```diff
@@ -208,13 +227,11 @@ protected function loadTranslatableData($locale = null)
...
+            return $value->attributes['locale'] === $locale;
+        });
 
-        $result = $obj ? json_decode($obj->attribute_data, true) : [];
+        $result = $obj ? json_decode($obj['attribute_data'], true) : [];
```
bennothommo: @massimomegistus Would you mind letting us know the reason for changing this line? The original line should've worked.

behaviors/TranslatableModel.php
mjauvin: The first() method's arguments should be ($key, $value) ...

ref. https://octobercms.com/docs/services/collections#method-first

mjauvin:
@massimomegistus I like your implementation!

behaviors/TranslatableModel.php
mjauvin: Actually, looks like the documentation referenced above is wrong the correct order for the callable arguments in the code is ($value, $key):

ref. `vendor/laravel/framework/src/Illuminate/Support/Arr.php` (line 175)

bennothommo: @mjauvin I was gonna say... I thought it was always $value, $key. :stuck_out_tongue:

bennothommo: Fixed in docs: https://github.com/octobercms/docs/commit/0ff464042c3630ff0026710c2f7376fd9685ea0e

behaviors/TranslatableModel.php
massimomegistus: @bennothommo there's no reason, In fact. That's a useless change.

bennothommo: @massimomegistus Feel free to restore it for us :)

massimomegistus:
@mjauvin thank you so much :)

massimomegistus:
Hello @w20k, yesterday I included the testGetTranslationValueEagerLoading in TranslatableModelTest

w20k:
@massimomegistus cool-cool, thanks! Already tested unit test ;) Works like a charm! 
Screenshot 2019-10-24 at 14 05 57

massimomegistus:
@w20k Great! Good news are the best!

behaviors/TranslatableModel.php
```diff
@@ -21,6 +21,25 @@
...
+    {
+        parent::__construct($model);
+
+        $model->morphMany['translations'] = [
+            'RainLab\Translate\Models\Attribute',
+            'name' => 'model'
+        ];
+    }
+
+    public function scopeTransWith($query)
```
LukeTowers: What's the point of this scope? Why wouldn't developers just use `with(['translations'])` by itself?

massimomegistus: Hi @LukeTowers, you are right. scopeTransWith could be chained to other scopes, but "with" option is much more flexible. I'll remove scopeTransWith without any regrets.

massimomegistus: @LukeTowers scopeTransWith method removed

mjauvin:
Looks good to me, I'll test this in real context and would be ready to merge unless someone else finds a problem.

mjauvin:
@hirisov can you test this PR and let us know if this addresses issue #502 ?

massimomegistus:
Hi guys, in here again. Do you think this PR could me merged in the next release, or you see any drawback? Thanks!

mjauvin:
Yes, we're only waiting for @hirisov to test this and confirm it's working and fixes his issue.

mjauvin:
@bunker-condenast can you test this PR and confirm this is working as expected?

bunker-condenast:
Hi @mjauvin , i'm already using it on two websites and it works fine

mjauvin:
@w20k, we're missing your approval on this, care to approve/disprove?

PR #509, opened 2019-08-30
Prevent from duplicate extending

tomaszstrojny:
While flushing model event listeners in `PluginTestCase.php` the plugin throws an exception that the class has already been extended. 

This fix will prevent the class from trying to extend it twice.

All in all, does anybody know why extending the class twice throws Exception instead of ignoring it and proceeding? :)

### DIFF
--- a/Plugin.php
+++ b/Plugin.php
@@ -45,8 +45,12 @@ class Plugin extends PluginBase
          */
         Page::extend(function($page) {
             $page->addDynamicProperty('translatable', ['title', 'description', 'meta_title', 'meta_description']);
-            $page->extendClassWith('RainLab\Translate\Behaviors\TranslatablePageUrl');
-            $page->extendClassWith('RainLab\Translate\Behaviors\TranslatablePage');
+            if (!$page->isClassExtendedWith('RainLab\Translate\Behaviors\TranslatablePageUrl')) {
+                $page->extendClassWith('RainLab\Translate\Behaviors\TranslatablePageUrl');
+            }
+            if (!$page->isClassExtendedWith('RainLab\Translate\Behaviors\TranslatablePage')) {
+                $page->extendClassWith('RainLab\Translate\Behaviors\TranslatablePage');
+            }
         });
 
         /*
@@ -54,8 +58,12 @@ class Plugin extends PluginBase
          */
         File::extend(function ($model) {
             $model->addDynamicProperty('translatable', ['title', 'description']);
-            $model->extendClassWith('October\Rain\Database\Behaviors\Purgeable');
-            $model->extendClassWith('RainLab\Translate\Behaviors\TranslatableModel');
+            if (!$model->isClassExtendedWith('October\Rain\Database\Behaviors\Purgeable')) {
+                $model->extendClassWith('October\Rain\Database\Behaviors\Purgeable');
+            }
+            if (!$model->isClassExtendedWith('RainLab\Translate\Behaviors\TranslatableModel')) {
+                $model->extendClassWith('RainLab\Translate\Behaviors\TranslatableModel');
+            }
         });
     }
 


mjauvin:
@czerwonyd this is due to this code block:

https://github.com/octobercms/library/blob/master/src/Extension/ExtendableTrait.php#L203-L209

czerwonyd:
@mjauvin yup, I know. But my question was more "why it's happening" more than "where's the throw instruction?".

I just think this should not throw any exceptions. I think it should be more gentle and just ignore duplicates. But maybe it makes sense I cannot see.

mjauvin:
Can you show your code in PluginTestCase.php that causes this behavior? I have not experienced this myself yet.

czerwonyd:
Sure, actually it's while writing tests with registering all the plugins, not only this one.

I have created a class `FullPluginTestCase` that is basically the same as the one suggested in the docs to register all plugins before running the test.

The problem occurs in the line `$pluginManager->unregisterAll()` with the following stacktrace:

```
Exception: Class System\Models\File has already been extended with October\Rain\Database\Behaviors\Purgeable

.../vendor/october/rain/src/Extension/ExtendableTrait.php:204                                                                                   
.../plugins/rainlab/translate/Plugin.php:57                                                                                                     
.../vendor/october/rain/src/Extension/ExtendableTrait.php:60
.../vendor/october/rain/src/Database/Model.php:68
.../vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/HasEvents.php:299
.../vendor/october/rain/src/Database/Model.php:209
.../tests/PluginTestCase.php:207
.../tests/PluginTestCase.php:100
.../plugins/initbiz/initdry/tests/classes/FullPluginTestCase.php:24
```

mjauvin:
@LukeTowers @bennothommo What's your opinion on this?

bennothommo:
@mjauvin I agree with @czerwonyd - you can technically double-include a trait in PHP, and it'll simply ignore any duplicate `use` calls, so given that behaviours in October are another implementation of traits, I think the same should apply.

I believe @LukeTowers is also checking with @daftspunk to determine the original intent of that check.

mjauvin:
@bennothommo thanks!
@LukeTowers, definitely keep us informed of what you find with @daftspunk

mjauvin:
@LukeTowers @daftspunk any new insight on the original intent of that check?

LukeTowers:
@mjauvin the only reason I can think of is to ensure that developers aren't expecting `extensionApplyInitCallbacks()` to be called the second time the extension is added (https://github.com/octobercms/library/blob/master/src/Extension/ExtendableTrait.php#L213) but I'm not convinced that's a good enough reason. I'll try to get @daftspunk to comment officially here.

daftspunk:
This likely is only a problem for `Page` (Halycon) and not for `File` (Eloquent). 

Note the database service provider clears the extensions every time it registers: https://github.com/octobercms/library/blob/8d96a7b239f8b30984ccde8dbfc525558d053197/src/Database/DatabaseServiceProvider.php#L33

Halycon does not clear the extensions on the model in the `register` method of its service provider:
https://github.com/octobercms/library/blob/8d96a7b239f8b30984ccde8dbfc525558d053197/src/Halcyon/HalcyonServiceProvider.php#L21-L28

I believe this could be the source of the problem

LukeTowers:
@tomaszstrojny could you look into @daftspunk's comment?

tomaszstrojny:
@LukeTowers, @daftspunk thanks for clarifying. It was definitely too deep for me to find :)

@mjauvin all in all, I guess it's time to merge the PR ;)

LukeTowers:
@mjauvin no...

LukeTowers:
@tomaszstrojny I believe that @daftspunk is suggesting that clearing the extensions on the Halcyon model in https://github.com/octobercms/library/blob/8d96a7b239f8b30984ccde8dbfc525558d053197/src/Halcyon/HalcyonServiceProvider.php#L35 could solve the root problem instead of having to apply all these isClassExtendedWith bandaid fixes.

@mjauvin please revert this commit as motivation to get the root cause dealt with.

daftspunk:
Yes please fix your tests and not the plugin.

PR #516, opened 2019-10-01
Add support for transOrderBy

aurelien-roy:
These changes add a new method enabling models to be sorted by an indexed translatable field.

Use case : Print a list of items sorted by their name where name is a translatable field. Of course, sorting won't be the same from one language to another.

Example :

```php
Product::transOrderBy('name')->get();
```

If a record hasn't be translated, the value from default locale will be used for sorting.


behaviors/TranslatableModel.php
```diff
@@ -48,6 +44,47 @@ public function scopeTransWhere($query, $index, $value, $locale = null, $operato
...
+
+    /**
+     * Joins the translatable indexes table to a query.
+     * This cannot be executed more than once.
+     * @param  Builder $query
+     * @param  string $locale
+     * @return Builder
+     */
+    protected function joinTranslateIndexesTable($query, $locale = null) {
+
```
LukeTowers: get rid of this extra line

LukeTowers:
@mjauvin is on vacation for the next few weeks so unless someone else reviews it it'll take a few weeks to get to this.

tests/unit/behaviors/TranslatableModelTest.php
```diff
@@ -101,4 +102,52 @@ public function testSetTranslationValue()
...
+
+        $obj->translateContext('fr');
+        $obj->name = 'Australie';
+        $obj->save();
+
+        $this->assertEquals(0, CountryModel::transWhere('name', 'Australie')->count());
+
+        Translator::instance()->setLocale('fr');
+        $this->assertEquals(1, CountryModel::transWhere('name', 'Australi')->count());
+
```
mjauvin: Is that a typo, should it be "Australie" instead of "Australi" ?

aurelien-roy: Indeed^^

mjauvin:
@LukeTowers looks good to me!

LukeTowers:
@mjauvin I don't see anything that stands out to me so if you've tested it and it works well feel free to merge when appropriate

mjauvin:
@LukeTowers @bennothommo I've tested this with the following OctoDock instance and it's working fine as far as I can see:

   https://d3f3b84cf746.octodock.com/backend

I also ran the phpunit tests within the instance without issues.

I'm ready to merge this if you have nothing further to add.

bennothommo:
@mjauvin Looks fine to me. It's new functionality, so its introduction shouldn't break anything.

aurelien-roy:
Thank you all for your time reviewing it!

behaviors/TranslatableModel.php
```diff
@@ -33,7 +33,7 @@ public function __construct($model)
 
     /**
      * Applies a translatable index to a basic query. This scope will join the index
-     * table and cannot be executed more than once.
+     * table and can be executed neither more than once, nor with scopeTransOrder.
```
LukeTowers: @mjauvin what does this mean?

mjauvin: Indeed... @aurelien-roy can you fix this comment so it makes sense?

PR #528, opened 2019-11-30
properly add translatable propertyto extended model

mjauvin:
make sure the translatable property gets added even if a plugin with higher precedence added it first


mjauvin:
the addDyamicProperty('translatable', [...]) call would have no effect if plugin loaded first also calls it.

PR #529, opened 2019-12-02
Slovenian language added

BlazOrazem:


lang/sl/lang.php
```diff
@@ -0,0 +1,59 @@
+<?php return [
+    'plugin'             => [
```
bennothommo: @BlazOrazem Please remove all the extra spacing between the first string and the arrow - there should only be one space to the left and right of the arrow.
```suggestion
    'plugin' => [
```
Please also do the same for all other lines below.

BlazOrazem: @bennothommo It's fixed.

bennothommo:
@BlazOrazem Perfect. Thanks for your contribution! :)

PR #532, opened 2019-12-18
add placeholders to isEmptyDataSet() verifications

### LINKED ISSUE #447: staticPage: Translated placeholders don't get saved without main content
### Expected behaviour
I create a new site in the static pages plugin and assign it to a layout with a placeholder. In the backend edit view I get two tabs: one for the main content and one for the placeholder.

grafik

Now I translate the placeholder without entering something for the main-content and click the save button. I expect the content for the placeholder gets saved.

### Actual behavior
The content of the placeholder doesn't get saved.

### Information about the Issue
Because the main content is not translated, the plugin does not create an additional file for the locale specific content. So the content of the placeholder field can not be saved anywhere.

### Steps to reproduce the behavior

1. Install the Static Pages Plugin and the Translate Plugin
1. Add at minimum two languages to the Translate Plugin
1. Add `{% placeholder MyPlaceHolder %}` to a layout
1. Add StaticPage component to this layout
1. Create a new Static Page using the above layout
1. Change to the MyPlaceHolder tab, enter some content and click save

mjauvin:
Fixes #447

Page would not save if only placeholders content was added/modified.

### DIFF
--- a/behaviors/TranslatableCmsObject.php
+++ b/behaviors/TranslatableCmsObject.php
@@ -166,7 +166,8 @@ class TranslatableCmsObject extends TranslatableBehavior
     protected function isEmptyDataSet($data)
     {
         return !array_get($data, 'markup') &&
-            !count(array_filter(array_get($data, 'viewBag', [])));
+            !count(array_filter(array_get($data, 'viewBag', []))) &&
+            !count(array_filter(array_get($data, 'placeholders', [])));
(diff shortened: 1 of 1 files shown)


mjauvin:
I've tested this, works as expected and resolves the bug.

@patrick-durrer can you confirm this resolves the issue on your end?

Issue #533, opened 2019-12-22
Mail Templates translation


mjauvin:
Content files can be translated by affixing the locale code to the file name. We need to apply the same technique to translate Mail template files. 

People are translating the mail templates themselves using the twig filters. This is a bad idea because when the mail is processed in a queue, it loses the locale setting.

mjauvin:
I think this could be implemented by hooking into the "mailer.beforeSend" event and modifying the $view parameter in the event handler

aurelien-roy:
Is this abandoned?

mjauvin:
Yes, it is closed.

PR #534, opened 2019-12-29
[WIP] localized mail templates

mjauvin:
use mail templates with locale suffix if it exists

fixes: #533

### DIFF
--- a/Plugin.php
+++ b/Plugin.php
@@ -34,6 +34,13 @@ class Plugin extends PluginBase
     public function register()
     {
         /*
+         * Adds language suffix to Mail templates.
+         */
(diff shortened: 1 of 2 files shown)


mjauvin:
- cannot use localized templates for privileged routes (e.g. /backend/auth)
- localized mail templates must be registered using "author.plugin::mail.template-LOCALE" (e.g. `rainlab.translate::mail.notification-fr`)

mjauvin:
@daftspunk are you sure the current locale would be available for a queued mail ?

daftspunk:
> @daftspunk are you sure the current locale would be available for a queued mail ?

It would appear not :(

mjauvin:
@daftspunk you're saying the locale is not available for a queued mail, can you provide a reproducible test for this?

PR #538, opened 2020-01-10
Fixes "Unparenthesized `a ? b : c ? d : e` is deprecated"

marcsoler:
Adds explicit parentheses on nested ternary operators for PHP 7.4 support

https://www.php.net/manual/en/migration74.deprecated.php#migration74.deprecated.core.nested-ternary


daftspunk:
Thanks @marcsoler !

PR #545, opened 2020-01-28
keep search term when filtering-out translated content

mjauvin:
This has been annoying me for a while. Before, everytime you clicked the "hide translated" checkbox, the current filtering based on search term was lost.


mjauvin:
Brain cramp, it's on locale selection that the search term is lost...

PR #549, opened 2020-02-09
Locale facade

mjauvin:
This change allows another plugin to improve on the Rainlab.Translate plugin by only extending its Locale model to add/change to its behavior.

Here's a short example on how to modify the base Locale model to always return all locales when running in development mode:

- add the following to your plugin's register() method:

```
 \App::extend('locale.class', function() {
   return new Classes\Locale
});
```

- create your Locale class that extends Rainlab.Translate Locale model:

```
<?php namespace Author\Plugin\Classes;

use App;

class Locale extends \RainLab\Translate\Models\Locale
{
    public function scopeIsEnabled($query)
    {   
        if (App::environment === 'dev') {
            // always return all configured locales
            return $query;
        }
        return parent::scopeIsEnabled($query);
    }   

    public static function listEnabled()
    {   
        if (App::environment() === 'dev') {
            // bypass cache
            return self::isEnabled()->order()->pluck('name', 'code')->all();
        }

        return parent::listEnabled();
    }
}
```

### DIFF
--- a/Plugin.php
+++ b/Plugin.php
@@ -1,5 +1,7 @@
 <?php namespace RainLab\Translate;
 
+use App;
+use Illuminate\Foundation\AliasLoader;
 use Lang;
 use Event;
 use Backend;
@@ -33,6 +35,13 @@ class Plugin extends PluginBase
 
     public function register()
     {
+        $aliasLoader = AliasLoader::getInstance();
+        $aliasLoader->alias('LocaleModel', 'RainLab\Translate\Facades\Locale');
+
+        App::singleton('locale.class', function () {
+            return new Models\Locale();
+        });
+
         /*
          * Defer event with low priority to let others contribute before this registers.
          */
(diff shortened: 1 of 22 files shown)


bennothommo:
@mjauvin Is there any particular reason you cannot simply extend the model in the normal fashion? ie. `\RainLab\Translate\Models\Locale::extend(function ($model) { ...`

LukeTowers:
I can see a need for this level of flexibility but I don't like the current approach. Facades seem like an overly heavy handed approach of dealing with this.

mjauvin:
@bennothommo ::extend() allows you to ADD methods but you can't modify/override an existing one. Am I the only one always bumping into this limitation?

mjauvin:
@LukeTowers I agree with you... any insightful suggestion on how best to approach this?

bennothommo:
@mjauvin I do bump into it on occasion, but seem to more often than not find a way around it :)

PR #554, opened 2020-02-13
escape Message translate params value to prevent XSS attack

### LINKED ISSUE #553: Plugin is not escaping the params value
consider this twig:

`{{ 'Welcome back :user'|_({user: user.firstname}) }}`

the value `user.firstname` is not escaped.

if `user.firstname` is something like `<script>alert('hi there')</script>`, the script will run

mjauvin:
Fixes #553


bennothommo:
@mjauvin This constitutes a breaking change, as it has been this way for ages. I would just be encouraging people to use the `|e` filter in Twig variables.

mjauvin:
@bennothommo this is what twig does already by default. So if you output the variable with twig directly like so `{{ user.firstName }}` it WILL be escaped. Shouldn't we have the same behavior here?

LukeTowers:
@bennothommo @mjauvin I would concur. I can't think of a single valid reason that someone would want parameters passed into the `| _()` filter to not be escaped, unless someone was already escaping that result with `| e`. However, as it hasn't been documented that `| _()` is unaffected by Twig's default escaping I think it's acceptable to make this change in the interest of security for our wider userbase at the expense of developers relying on the current undocumented innate insecurity of `| _()`.

bennothommo:
@mjauvin Normal output of variables through Twig is escaped, but as soon as a filter is involved, it is passed through *unescaped* to the filter, and it then becomes the responsibility of the filter to escape the result as necessary. October allows filters (both in-built and pluign-provided) to run in HTML mode here: https://github.com/octobercms/october/blob/master/modules/system/classes/MarkupManager.php#L276

@LukeTowers while you are right in that it is unlikely, there are cases for HTML being added to the parameters for translation, like for example, super or sub-script characters for a title.

Eg. `{{ 'Title: :title', _({title: "E = MC<sup>2</sup>"}) }}`

Given that Twig is not only used for just user-provided variables, but also for default markup in plugin components where data might be provided by something already handling escaping, I don't think it's safe to just blanket escape everything.

mjauvin:
Hmmm, interesting counter argument @bennothommo

LukeTowers:
Hmm, I missed the fact that all custom filters registered in October through the plugin registration files bypass the escape protection provided by Twig by default. That seems a bit like a foot gun to me. What would you recommend as the solution then? Should we just update the docs (both https://octobercms.com/docs/plugin/registration#extending-twig and the plugin docs) to recommend using `| e` after any custom filter that you use that will have user input in it? Perhaps @daftspunk can take a look at this in a week when he's back as well.

bennothommo:
@LukeTowers I think that's probably the only thing we can do that won't constitute a BC break for that functionality.

mjauvin:
hmm, what about escaping only if a "<script>" tag is found in the string?

bennothommo:
@mjauvin that might be too brittle a check - Javascript can be run in different ways, for example, through event handler attributes like `onload`.

mjauvin:
@bennothommo true, so I guess we'll just leave the code as it is, and add a note to documentation.

LukeTowers:
Still waiting for @daftspunk's input on this so we can leave it open for now.

mjauvin:
@daftspunk did you have a chance to look into this one?

daftspunk:
This is clearly an oversight in the original design, since escaped variables is the common/expected functionality. This change is fine to push through so long as everyone is made aware of the modified behavior, but importantly a workaround is in place for those that intend to have unescaped variables. What is the solution for these people? Is it to use `|raw`?

Eg. {{ 'Title: :title', _({title: "E = MC<sup>2</sup>"})|raw }}

Once a confirmed solution is in place to retain the original functionality (for the unlikely few that may want it), this can be merged and we should include appropriate guidance in the release notes.

README.md
```diff
@@ -46,6 +46,10 @@ Message or string translation is the conversion of adhoc strings used throughout
 
     {{ 'Hello :name!'|_({ name: 'Friend' }) }}
 
+Translated content is escaped. If you want to use HTML content in your translated string, you can use the `raw` filter.
```
LukeTowers: @mjauvin did you verify that this works?

mjauvin: I just tested this and it does not work... there is a typo, should have been:
```
{{ 'Title:  :title'| _({title: "E = MC<sup>2</sup>"})|raw }}
```

But still does not work...

mjauvin:
I tested an alternative to escaping the string... instead using Html::clean() on the string... it will let regular HTML through, but will clean the string of potential XSS. see below:
```
diff --git a/models/Message.php b/models/Message.php
index 4f892e9..731eefe 100644
--- a/models/Message.php
+++ b/models/Message.php
@@ -205,7 +205,7 @@ class Message extends Model
         $msg = static::get($messageId, $locale);
    
         $params = array_build($params, function($key, $value){
-            return [':'.$key, e($value)];
+            return [':'.$key, html_clean($value)];
         });
 
         $msg = strtr($msg, $params);

```

mjauvin:
So with this, the following:
```
{{ 'Title: :title'|_({title: '<script>window.location = "http://google.com"</script>'}) }}
```
Renders as:
```
Title: window.location = "http://google.com"
```

But the following:
```
{{ 'Title: :title'|_({title: "E = MC<sup>2</sup>"}) }}
```

Renders as:
```
Title: E = MC2 (with "2" as superscript of course)
```

README.md
LukeTowers: That's because the escaping is done in the filter itself. We could potentially take an optional second argument (default true) to escape the output

mjauvin: The method already has MANY arguments, though:
```
    public static function trans($messageId, $params = [], $locale = null)
```

daftspunk:
`Html::clean` will be very slow by comparison, I think you should introduce a verbose `transRaw` function to help people that need the old functionality

```
{{ 'Title: :title'|transRaw({title: "E = MC<sup>2</sup>"}) }}
```

mjauvin:
@daftspunk we currently have two twig filters using Message::trans:

- _() -> calls translateString()
- __() -> calls translatePlural()

Should we introduce two new filters:

- transRaw() or _raw()
- transRawPlural() or __raw()

?

mjauvin:
The other option is adding a 4th parameter to the filters and Message::trans like so:
```
public function trans($string, $params = [], $locale = null, $raw = false)...
public function translateString($string, $params = [], $locale = null, $raw = false)...
public function translatePlural($string, $count = 0, $params = [], $locale = null, $raw = false)...
```
So a raw call in twig would become:
```
_("My String", [], null, true)
```

daftspunk:
@mjauvin Don't double down by adding more args to `_`, just introduce `transRaw` for the people that need it. We assert that nobody does (so no 4th arg needed), but just in case someone does, we are there for them

multiwebinc:
Just as a comment, I encountered this issue when including links in a string, which I wouldn't think would be *that* uncommon in websites.

```
{{ 'Please go to our :contact_page.'|_({contact_page: '<b><a href="' ~ 'contact'|page ~ '">' ~ 'contact page'|_ ~ '</a></b>'}) }}
```

Since this is a breaking change, would it not make more sense to change it in two steps? First introduce `transRaw` and `transRawPlural`, and on a future update change the default behavior?

LukeTowers:
@multiwebinc I don't think that would help much, but we should do a better job warning users about the change. 

@mjauvin can you release a 1.7.0 !!! update that warns about the change? Just make the change to the version file in a PR and I'll review it and then update the marketplace.

mjauvin:
@LukeTowers does it have to be version 1.7.0 ? My next milestone is 1.6.11

LukeTowers:
@mjauvin technically it should be 2.0.1, it's a bit of a breaking change. Feel free to rename 1.6.11 to 1.7.0 and move any un-finished issues to 1.7.1. It's not quite semantic versioning but it's a bit more accurate than changing just the minor version number.

mjauvin:
@LukeTowers do you actually want me to create a release or just a PR with the version.yaml update?

LukeTowers:
@mjauvin just make it a new release.

mjauvin:
@LukeTowers do you still want to approve PR #573 before I cut the release?

PR #556, opened 2020-02-13
Fix error with nested fields default locale value

Trysystems:


mjauvin:
Can you please reference the issues this would fix?

Trysystems:
Sorry, it's my first PR, I don't know how to do this)

mjauvin:
Although this might fix the problem you are having, this does not fix the problem at the root. The problem lies deeper and affects more than MLControl based fields.

I am closing this and reopening PR associated with #303 to try and fix the problem at the root.

mjauvin:
This is a very simple fix which seems to work fine to address the issue without breaking anything.

@osmanzeki would you ming testing this?

mjauvin:
Replaces #322

mjauvin:
This can be tested here:

https://353983ad8721.octodock.com/backend

user/pass = admin

I modified the Playground's Country model "name" field to be casts of type array and the name[description] field has been added to the translatable fields.

Feel free to experiment.

Issue #376, opened 2018-05-28
Theme Data Translation


raquelegluu:
I'm trying to translate theme customization fields extending ThemeData like I see on issue https://github.com/rainlab/translate-plugin/issues/279 for extend mail templates, but it isn't working.

In my boot method of my plugin I put this:
```
ThemeData::extend(function($model) {
    $model->implement[] = '@RainLab.Translate.Behaviors.TranslatableModel';
    $model->addDynamicProperty('translatable', ['website_data']);
});
```

The field 'website_data' is configured in theme.yaml
```
 website_data:
    type: text
    tab: Info
    label: Data Tanslatable
```

The field is printed correctly but without translation option.

Is it possible to translate this fields?? Thanks

LukeTowers:
@raquelegluu it's possible that's not supported as all of those fields are actually collected and stored in the `data` model attribute: https://github.com/octobercms/october/blob/master/modules/cms/models/ThemeData.php#L56-L63 which might not play nice with this plugin. Feel free to investigate further.

raquelegluu:
Thanks @LukeTowers, finally I saw it and solved it with my own event to read all translatable fields from themeData and creating my own behavior to manage data before save it.

LukeTowers:
@raquelegluu awesome, glad to hear it! If you can make a PR to provide that ability to other people that would be fantastic!

iboved:
@raquelegluu How did you do this? Can you give an example of how to make ThemeData fields translatable?
Thank you in advance.

AdrienAdB:
Hi, working on same feature. Did anyone find a solution? @raquelegluu ?

mjauvin:
Closing as it has been over a month since any activity on this occurred and we are trying to figure out what issues are still relevant. If this is still something that you would like to see through to fruition please respond and we can get the ball rolling.

robinbonnes:
Needed this too, fixed it by adding the following code to the `plugin.php` boot function in any existing plugin, or create a new plugin:

```
<?php namespace Acme\Translate;

class Plugin extends \System\Classes\PluginBase
{
    public function pluginDetails() 
    {
        return [
            'name'        => 'Translate Theme',
            'description' => 'With this plugin you can translate theme data.'
        ];
    }

    public function boot() 
    {
        if(!class_exists('\RainLab\Translate\Behaviors\TranslatableModel'))
            return;

        \Cms\Models\ThemeData::extend(function ($model) { 
        
            // Implement translatable model
[...]

PR #559, opened 2020-02-27
Fix for rainlab/translate-plugin#376

GinoPane:
Fields are not marked as translatable by default.
It is a theme author responsibility to configure required fields via theme's config and mark fields as `translatable`.

For example:

```
tabs:
  fields:

    ####################################################
    # Tab: Info
    ####################################################

    website_name:
      tab: Info
      label: Website Name
      type: text
      default: Octaskin
      translatable: true
```

And the result is:

image

### DIFF
--- a/Plugin.php
+++ b/Plugin.php
@@ -69,14 +69,23 @@ class Plugin extends PluginBase
         /*
          * Add translation support to theme settings
          */
-        ThemeData::extend(function ($model) {
+        ThemeData::extend(static function ($model) {
             $model->extendClassWith(TranslatableModel::class);
 
             if (!$model->propertyExists('translatable')) {
                 $model->addDynamicProperty('translatable', []);
             }
 
-            $model->bindEvent('model.afterFetch', function() use ($model) {
+            // For some reason translatable property may appear in model's attributes
+            // and this is undesired behavior. If it happens we need to unset the attribute to prevent
+            // 'Unexpected type of array when attempting to save attribute "translatable"' exception
+            $model->bindEvent('model.saveInternal', static function() use ($model) {
+                if (isset($model->attributes['translatable']) && is_array($model->attributes['translatable'])) {
+                    unset($model->attributes['translatable']);
+                }
+            });
+
+            $model->bindEvent('model.afterFetch', static function() use ($model) {
                 foreach ($model->getFormFields() as $id => $field) {
                     if (!empty($field['translatable'])) {
                         $model->translatable[] = $id;
--- a/Plugin.php
+++ b/Plugin.php
@@ -10,7 +10,6 @@ use System\Classes\PluginBase;
 use RainLab\Translate\Models\Message;
 use RainLab\Translate\Classes\EventRegistry;
 use RainLab\Translate\Classes\Translator;
-use RainLab\Translate\Behaviors\TranslatableModel;
 
 /**
  * Translate Plugin Information File
@@ -70,20 +69,12 @@ class Plugin extends PluginBase
          * Add translation support to theme settings
          */
         ThemeData::extend(static function ($model) {
-            $model->extendClassWith(TranslatableModel::class);
-
             if (!$model->propertyExists('translatable')) {
                 $model->addDynamicProperty('translatable', []);
             }
 
-            // For some reason `translatable` property may appear in model's attributes
-            // and this is an undesired behavior. If it happens we need to unset the attribute to prevent
-            // 'Unexpected type of array when attempting to save attribute "translatable"' exception
-            $model->bindEvent('model.saveInternal', static function() use ($model) {
-                if (isset($model->attributes['translatable']) && is_array($model->attributes['translatable'])) {
-                    unset($model->attributes['translatable']);
-                }
-            });
+            $model->extendClassWith('October\Rain\Database\Behaviors\Purgeable');
+            $model->extendClassWith('RainLab\Translate\Behaviors\TranslatableModel');
 
             $model->bindEvent('model.afterFetch', static function() use ($model) {
                 foreach ($model->getFormFields() as $id => $field) {
--- a/Plugin.php
+++ b/Plugin.php
@@ -76,8 +76,8 @@ class Plugin extends PluginBase
                 $model->addDynamicProperty('translatable', []);
             }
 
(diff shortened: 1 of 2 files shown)


Plugin.php
```diff
@@ -64,6 +66,25 @@ public function register()
...
+        ThemeData::extend(function ($model) {
+            $model->extendClassWith(TranslatableModel::class);
+
+            if (!$model->propertyExists('translatable')) {
+                $model->addDynamicProperty('translatable', []);
+            }
+
+            $model->bindEvent('model.afterFetch', function() use ($model) {
+                foreach ($model->getFormFields() as $id => $field) {
+                    if (!empty($field['translatable'])) {
```
mjauvin: You assume the translatable property is true by default. This is a boolean, we should be able to set it to false and expect it NOT to be translatable.

GinoPane: Most themes won't have this property in their configs, so by default we should expect that it does not even exist. But if it exists and it is true (what `!empty` guarantees) we add the corresponding field into translatable list.

mjauvin: Yes, but down the line, if a field has the translatable property, someone would expect that changing its value to false would disable the translation for the field

GinoPane: Yes, it is exactly how it works. 
If no `translatable` or `translatable=false` - no action, if `translatable=true` the field is picked up for translation.

mjauvin: Oh, sorry, didn't realize `empty()` treated false value as empty.

mjauvin:
@GinoPane I tried to test this PR using the following instance:

https://88f9e4ea8aac.octodock.com/backend

(user/pass = admin)

I have the following `theme.yaml` file:
```
name: demo
description: demo theme
author: OctoberCMS
homepage: 'https://octobercms.com'
code: ''
form:
  tabs:
    fields:
      website_name:
        tab: Info
        label: Website Name
        type: text
        default: Your website name
        translatable: true
```

And I get the following error when visiting the Cusomize Theme page:

`Form record ID has not been specified.`

GinoPane:
@mjauvin it's really strange, now it is working, though code is the same. I tried to debug a little, commented plugin changes, loaded the page, after that un-commented changes and it worked 🤷‍♂ 

image

mjauvin:
Well, you probably did something that got cached or something. I created a new instance, added the same theme.yaml and I get the same error again:

https://cfc0174a7a29.octodock.com/backend

GinoPane:
@mjauvin Hm, it seems to be happening when cms_theme_data record is missing for the theme. For any reason the system cannot create the record.

It's interesting 🤔

mjauvin:
How did you get it created the first time?

GinoPane:
@mjauvin when I commented the code, the record was created. I don't fully understand the issue, but for any reason dynamic `translatable` property got into model's attributes and this exception was thrown: `Unexpected type of array when attempting to save attribute "%s", try adding it to the $jsonable property.`

This exception was caught in ThemeData::forTheme method and therefore the generic error message was shown on the front end.

Anyway, I committed a fix which should be working fine.

Plugin.php
```diff
@@ -69,14 +69,23 @@ public function register()
         /*
          * Add translation support to theme settings
          */
-        ThemeData::extend(function ($model) {
+        ThemeData::extend(static function ($model) {
             $model->extendClassWith(TranslatableModel::class);
```
mjauvin:
```suggestion
            $model->extendClassWith('October\Rain\Database\Behaviors\Purgeable');

```

Plugin.php
```diff
@@ -69,14 +69,23 @@ public function register()
...
 
-            $model->bindEvent('model.afterFetch', function() use ($model) {
+            // For some reason translatable property may appear in model's attributes
+            // and this is undesired behavior. If it happens we need to unset the attribute to prevent
+            // 'Unexpected type of array when attempting to save attribute "translatable"' exception
+            $model->bindEvent('model.saveInternal', static function() use ($model) {
+                if (isset($model->attributes['translatable']) && is_array($model->attributes['translatable'])) {
+                    unset($model->attributes['translatable']);
+                }
+            });
```
mjauvin:
```suggestion
```

mjauvin: This is not needed when adding the Purgeable Behavior

GinoPane: Yep, thanks! It's interesting, because I tried this before I worked on my specific solution and this trait did not help. Maybe anything else was wrong at the moment 😄 Anyway, now it's better and seems to be working.

Plugin.php
```diff
@@ -64,6 +65,26 @@ public function register()
             $model->extendClassWith('RainLab\Translate\Behaviors\TranslatableModel');
         });
 
+        /*
+         * Add translation support to theme settings
+         */
+        ThemeData::extend(static function ($model) {
```
mjauvin: Why the "static" keyword here?

Plugin.php
```diff
@@ -64,6 +65,26 @@ public function register()
...
+            if (!$model->propertyExists('translatable')) {
+                $model->addDynamicProperty('translatable', []);
+            }
+
+            $model->extendClassWith('October\Rain\Database\Behaviors\Purgeable');
+            $model->extendClassWith('RainLab\Translate\Behaviors\TranslatableModel');
+
+            $model->bindEvent('model.afterFetch', static function() use ($model) {
```
mjauvin: and here?

Plugin.php
mjauvin: The thing is hard to get right, the translatable property needs to be added first, then you need to extend with Purgeable and THEN you need to extend with TranslatableModel...

If any of the above is reversed, it won't work...

Plugin.php
mjauvin: Otherwise, I'm ready to merge this.

Plugin.php
GinoPane: Yes, definitely. I had extension at first and property added afterwards and that's why it did not work.

Plugin.php
GinoPane: It is just a small optimisation. Static closures won't utilize the outer scope and therefore they are faster:

image

mjauvin: nice, didn't know that.

PR #560, opened 2020-03-11
Register asset bundle

mjauvin:
- add registerBundle() calls to allow compiling LESS files into CSS
- make sure multi-lingual input form controls have padding-right of 44px

### DIFF
--- a/assets/less/multilingual.less
+++ b/assets/less/multilingual.less
@@ -24,6 +24,9 @@
     }
 
     &.field-multilingual-text {
+        input.form-control {
+            padding-right:44px
+        }
         .ml-btn {
             right: 4px;
             top: 50%;
@@ -110,7 +113,10 @@
             }
         }
     }
-    .field-multilingual input.form-control {
-        padding-right:44px
+}
+
+.help-block.before-field + .field-multilingual.field-multilingual-textarea {
+    .ml-btn {
+        top:-41px;
     }
 }
--- a/assets/less/multilingual.less
+++ b/assets/less/multilingual.less
@@ -24,10 +24,6 @@
     }
 
     &.field-multilingual-text {
-        input.form-control {
-            padding-right: 44px;
-        }
-
         .ml-btn {
             right: 4px;
             top: 50%;
@@ -39,12 +35,12 @@
     &.field-multilingual-textarea {
         .ml-btn {
             right: 11px;
-            top: 5px;
+            top: -29px;
             text-align: right;
         }
 
         .ml-dropdown-menu {
-            top: 39px;
+            top: 1px;
             right: 1px;
         }
     }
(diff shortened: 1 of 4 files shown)


assets/less/multilingual.less
```diff
@@ -24,7 +24,7 @@
     }
 
     &.field-multilingual-text {
-        .form-control {
+        input.form-control {
```
bennothommo: @mjauvin `form-control` elements aren't always inputs (eg. textareas), so I think it was set to just `.form-control` for that reason. From your Slack convo, you are wanting to fix just the elements in the "fancy" layout, so I would instead add a new block within here somewhere: https://github.com/rainlab/translate-plugin/blob/1b1792caebd0f2904f666b8c2ece0c3f3c6bebc7/assets/less/multilingual.less#L108

bennothommo: Whoops, I see that the `.field-multilingual-text` class above it would only apply to text fields. Nonetheless, it's probably better to make the change within the `fancy-layout` block.

mjauvin: I'll make the change, makes sense.

mjauvin:
I noticed the language selector for textarea is problematic because of the scrollbar that appears dynamically once the content overflows:

before scrollbar appears (language selector can be obscured by text):
Screenshot from 2020-03-12 10-12-18

after scrollbar appears:
Screenshot from 2020-03-12 10-14-11


I propose we change the location of the language selector OUTSIDE of the field to cope for this, see screenshot below:

Screenshot from 2020-03-12 10-06-51


What do you guys think?

LukeTowers:
@mjauvin sounds good to me, just make sure it's on the same horizontal line as the field label

bennothommo:
@mjauvin You'll need to check the placement of the language selector when someone uses the `commentAbove` field option. If it stays there, then that's fine.

mjauvin:
@bennothommo good point. Look at the following screenshot of a textarea with commentAbove:

Screenshot from 2020-03-12 22-20-15

Is that what you would expect?

bennothommo:
@mjauvin If possible, it should line up with the lowest element in the "label" portion of the field, so in this case, it should probably line up with the comment. But otherwise, it looks fine to me.

mjauvin:
This is how it looks now with commentAbove:

Screenshot from 2020-03-13 01-24-52

assets/less/multilingual.less
mjauvin: could not make this change since fancy layout is not used for every multilang text fields.

bennothommo: @mjauvin Does the same issue occur with text fields not within a fancy layout though?

mjauvin: @bennothommo you're right, it doesn't with the initial CSS. All good now!

Issue #563, opened 2020-03-20
Changing non-alphanumeric characters does not change the string


multiwebinc:
Let's say, for example, I have the following string in one of my views:

```html
<p>{{ 'Choose one of the following:'|_ }}</p>
```

Then I later change my mind and decide to move the `:` outside the translation string:

```html
<p>{{ 'Choose one of the following'|_ }}:</p>
```

I then do a re-scan of the translate messages, however the string is still listed as `Choose one of the following:` with the `:` within the string, and when the translators see this, they will assume they need to include the `:` as well.

I see that in the `rainlab_translate_messages` table, the `code` column used looks like it removes all non-alphanumeric characters and replaces the spaces with a dot. Wouldn't it be more sensible to use a hash of the string instead of this? That way a single whitespace character change would trigger the whole string to be changed.

mjauvin:
Yes, this is a known limitation. You need to delete the translation message and rescan.

What you suggest would wirk, but would cause major backward compatibility issues.

multiwebinc:
@mjauvin What major backward-compatibility issues would it cause? Besides using it for an index to quickly find strings internally in the plugin, do other plugins somehow access these strings for another purpose?

mjauvin:
@multiwebinc Anyone with the plugin currenlty installed would have major problems when updating the plugin to a version using a hash for the code since all their existing translations would be lost.

We'd need to do a migration but this could potentially create havoc.

mjauvin:
FYI, I submitted a PR to allow overriding the method used to generate the messageCode.

ref, #564

multiwebinc:
> We'd need to do a migration but this could potentially create havoc.

This is exactly what I was thinking as well. As long as no external plugins rely explicitly on the storage method used for the Translate plugin, run a migration on next update that simply updates all current codes to hashes.

Another option would be to select rows where the row is either a hash or the result of the current function. Of course, that adds slightly more computational power, but might be an option to allow the current code calculation method to be deprecated over several releases before it's finally removed.

mjauvin:
@daftspunk what's your take on this, should we transition to md5 hash as we discussed before?

daftspunk:
This has been fixed in October CMS v3.1 with a new translation key format.

- Link to the [upgrade guide](https://github.com/rainlab/translate-plugin/blob/master/UPGRADE.md)
- Link to [multisite documentation](https://docs.octobercms.com/3.x/cms/resources/multisite.html)

[...]

PR #564, opened 2020-03-20
add rainlab.translate.messageCode event

mjauvin:
Proposal to introduce an event to override how the messageCode is generated.

Usage example:
```
Event::listen('rainlab.translate.messageCode', function ($messageId) {
   return md5($messageId);
});
```

### DIFF
--- a/models/Message.php
+++ b/models/Message.php
@@ -1,5 +1,6 @@
 <?php namespace RainLab\Translate\Models;
 
+use Event;
 use Str;
 use Lang;
 use Model;
@@ -282,6 +283,10 @@ class Message extends Model
      */
     protected static function makeMessageCode($messageId)
     {
+        if ($result = Event::fire('rainlab.translate.messageCode', [$messageId], true)) {
+            return $result;
+        }
+
         $separator = '.';
 
         // Convert all dashes/underscores into separator


LukeTowers:
@mjauvin how often in a request does this method get called? Additionally it should probably be `rainlab.translate.makeMessageCode`

mjauvin:
@LukeTowers I would say as many times as there are translation calls.

If you're worried about efficiency, I would think this is way more efficient than the 3 calls to preg_replace() that follows...

LukeTowers:
@mjauvin try installing the Debugbar plugin and enabling the event listener in the config (you'll also need to apply https://github.com/barryvdh/laravel-debugbar/pull/1014 to get it working) and you'll see what sort of impact frequently called events can have.

mjauvin:
@LukeTowers So you're saying events are not efficient and should not be used too often in a page render?

LukeTowers:
Events by themselves are not inherently inefficient, it's just that we should always carefully consider the addition of any event, since it represents a potential liability for us as maintainers in that once we introduce an event we can't ever remove it. Additionally, they can represent a potential footgun for developers if they're fired in a place that would be called multiple times per request. These are the factors that we have to carefully consider anytime we're reviewing the addition of any new events.

Are there any other ways that your desired end result could be achieved?

mjauvin:
Of course, we could return the md5 of the messageId directly, but that would require a migration for all existing Message records.

Not sure how safe that would be. Using the MD5 hash of the Message Id would be a much better choice and would solve many problems related to the way this is currently encoded.

multiwebinc:
As I [mentioned](https://github.com/rainlab/translate-plugin/issues/563#issuecomment-601984652) in my bug report, both methods could be supported simultaneously for a while, in case, for example, people update the plugin, but don't perform the migration right away. Or are they performed automatically when plugins are updated? Also this could be an issue if, for whatever reason, other plugins rely explicitly on the current message code generation function, however that doesn't seem likely to me since `makeMessageCode()` is `protected` and the two times in that class that this method it is called, the actual format of the code returned is irrelevant.

bennothommo:
@mjauvin I know you like to rail on regular expressions, but I'm curious, do you have any hard numbers on the speed of using the old system with the regular expressions over using an MD5 hash - I would imagine they would be comparable?

mjauvin:
> @mjauvin I know you like to rail on regular expressions, but I'm curious, do you have any hard numbers on the speed of using the old system with the regular expressions over using an MD5 hash - I would imagine they would be comparable?

@bennothommo In this particular case, I was just making the point that 3 regex calls was not better than one event being called.

But the idea about replacing all this with an MD5 hash is that it would help achive the following:

- a code for "my msg id:" or "my msg id" or "my:msg:id:" would all be different (now they are the same)
- a code for a LONG message id would always be 32 characters and would not have be truncated

mjauvin:
I'll close this and submit a PR to use md5 hash to generate the messageCode.

multiwebinc:
@mjauvin Updating the current codes is pretty trivial:

```php
foreach (Message::all() as $message) {
    $default_message = $message->message_data['x'];
    $message->code = md5($default_message);
    $message->save();
}
```
Then in `Message.php` just replace `makeMessageCode` with `md5` and remove the `makeMessageCode` method. I tested it out and it works fine for me. Re-scanning actually found a couple of other otherwise hidden messages that probably varied by case or punctuation.

Issue #525, opened 2019-11-06
Allow purging messages to keep existing translations


multiwebinc:
When you do a scan for messages and check the "purge all messages first", this deletes everything, including existing translations, before re-adding the strings the scanner finds. And if you don't check the "purge all messages" box, any time you update a translation in the source files will create an extra message to be translated.

You are therefore left with two choices; check the box and have to re-translate everything again, or not check that box and end up with a bunch of ghost strings that are not found anywhere in the project. Neither of these cases is ideal and both will necessarily cause extra work for translators.

Here are my two suggestions:

- Add an option to purge only the strings that are not still in the project.
- Add a column to the list view to indicate that the string couldn't be found in the project anymore to give the translator an idea which strings can be safely deleted.

mjauvin:
@multiwebinc any chance you could submit a PR for the proposed suggestion(s) ?

PR #565, opened 2020-03-25
Add option to purge only deleted messages

multiwebinc:
Related bug report: #525 

I've added another checkbox in the scanner window that when checked will purge only the messages that the scanner was unable to find. For the life of me I couldn't figure out how to add a column to the list view to indicate that the string couldn't be found in the project anymore to give the translator an idea which strings can be safely deleted.

~~It is also missing a migration to add a `boolean` column to `rainlab_translate_messages` called `found`. I'm not exactly sure how that should be handled since you need to increment the version.~~

### DIFF
--- a/classes/ThemeScanner.php
+++ b/classes/ThemeScanner.php
@@ -33,19 +33,13 @@ class ThemeScanner
      */
     public function scanForMessages()
     {
-        $this->setMessagesAsNotFound();
+        // Set all messages initially as being not found. The scanner later
+        // sets the entries it finds as found.
+        Message::query()->update(['found' => false]);
+
         $this->scanThemeConfigForMessages();
         $this->scanThemeTemplatesForMessages();
         $this->scanMailTemplatesForMessages();
-    }
-
-    /**
-     * Sets all current messages as not found
-     * @return void
-     */
-    public function setMessagesAsNotFound()
-    {
-        Message::query()->update(['found' => 0]);
     }
 
     /**
--- a/classes/ThemeScanner.php
+++ b/classes/ThemeScanner.php
@@ -33,9 +33,19 @@ class ThemeScanner
      */
     public function scanForMessages()
     {
+        $this->setMessagesAsNotFound();
         $this->scanThemeConfigForMessages();
         $this->scanThemeTemplatesForMessages();
         $this->scanMailTemplatesForMessages();
+    }
+
+    /**
+     * Sets all current messages as not found
+     * @return void
+     */
+    public function setMessagesAsNotFound()
+    {
+        Message::query()->update(['found' => 0]);
     }
 
     /**
--- a/controllers/messages/config_table.yaml
+++ b/controllers/messages/config_table.yaml
@@ -13,3 +13,6 @@ columns:
         title: From
     to:
         title: To
+    found:
+        title: Found
+        width: 100px
--- a/controllers/messages/config_table.yaml
+++ b/controllers/messages/config_table.yaml
@@ -14,5 +14,5 @@ columns:
     to:
         title: To
     found:
-        title: Found
-        width: 100px
+        title: Scan errors
+        width: 150px
--- a/models/Message.php
+++ b/models/Message.php
@@ -86,8 +86,8 @@ class Message extends Model
         if (!$message) {
             unset($data[$locale]);
         }
-
         $this->message_data = $data;
+
         $this->save();
     }
 
@@ -126,6 +126,7 @@ class Message extends Model
         if (!$item->exists) {
             $data = [static::DEFAULT_LOCALE => $messageId];
             $item->message_data = $item->message_data ?: $data;
+
             $item->save();
         }
 
@@ -162,6 +163,8 @@ class Message extends Model
             $locale = static::DEFAULT_LOCALE;
         }
 
+        $existingIds = [];
+
         foreach ($messages as $code => $message) {
             // Ignore empties
             if (!strlen(trim($message))) {
@@ -171,7 +174,7 @@ class Message extends Model
             $code = self::makeMessageCode($code);
 
             $item = static::firstOrNew([
-                'code' => $code
+                'code' => $code,
             ]);
 
             // Do not import non-default messages that do not exist
@@ -183,8 +186,7 @@ class Message extends Model
 
             // Do not overwrite existing translations
             if (isset($messageData[$locale])) {
-                $item->found = 1;
-                $item->save();
+                $existingIds[] = $item->id;
                 continue;
             }
 
@@ -192,8 +194,11 @@ class Message extends Model
 
             $item->message_data = $messageData;
             $item->found = 1;
+
             $item->save();
         }
+
+        self::whereIn('id', $existingIds)->update(['found' => 1]);
     }
 
     /**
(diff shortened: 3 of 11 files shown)


mjauvin:
Thanks a lot!
 I'll take a look on how we could add those columns to the messages list.

mjauvin:
I had a quick look and I didn't see where the messages actually get purged... what am I missing?

multiwebinc:
I missed uploading the controller.

models/Message.php
```diff
@@ -183,12 +183,15 @@ public static function importMessageCodes($messages, $locale = null)
 
             // Do not overwrite existing translations
             if (isset($messageData[$locale])) {
+                $item->found = 1;
+                $item->save();
```
LukeTowers: Is there a way we can update all existing messages with the found status in one DB query instead of one per message? That will be quite a performance impact on sites with large translation bases.

classes/ThemeScanner.php
```diff
@@ -33,11 +33,21 @@ public static function scan()
      */
     public function scanForMessages()
     {
+        $this->setMessagesAsNotFound();
```
LukeTowers: Inline this method as 
```suggestion
        // Set all messages as not found to be updated one by one.
        Message::query()->update(['found' => false]);
```

models/Message.php
mjauvin: It's only done when manually scanning the messages, but yeah, if we could somehow update with one query, that would be ideal.

mjauvin: One way to improve this would be to append Message records ids into an array like:
```
if (isset($messageData[$locale])) {
   $existingIds[] = $item->id;
   continue;
}   
```
And run the following query after the loop:
```
Message::whereIn('id', $existingIds)->update(['found' => 1]);
```

mjauvin:
@multiwebinc for the migration, just create the `update_messages_table.php` migration file and I'll take care of the version.yaml file.

mjauvin:
update_messages_table.php:
```
<?php namespace RainLab\Translate\Updates;

use Schema;
use October\Rain\Database\Updates\Migration;

class UpdateMessagesTable extends Migration
{
    public function up()
    {   
        Schema::table('rainlab_translate_messages', function($table)
        {   
            $table->boolean('found')->default(1);
        });
    }       

    public function down()
    {   
        Schema::table('rainlab_translate_messages', function($table)
        {   
            $table->dropColumn('found');
        });
    }
}
```

models/Message.php
```diff
@@ -204,26 +212,6 @@ public static function trans($messageId, $params = [], $locale = null)
...
-     * Looks up and translates a message by its string WITHOUT escaping params.
-     * @param  string $messageId
-     * @param  array  $params
-     * @param  string $locale
-     * @return string
-     */
-    public static function transRaw($messageId, $params = [], $locale = null)
-    {
-        $msg = static::get($messageId, $locale);
-
```
mjauvin: The above should not be removed.

mjauvin:
LGTM! I ran a few tests and it seems to be working as expected!

multiwebinc:
@mjauvin How hard would it be to put a small exclamation mark type icon on the message list page for translations marked as not found? I couldn't easily figure that out the other day.

mjauvin:
The markup for this is rather *alien* to me, couldn't figure it out either.

models/Message.php
```diff
@@ -86,8 +86,8 @@ public function toLocale($locale = null, $message)
         if (!$message) {
             unset($data[$locale]);
         }
-
         $this->message_data = $data;
+
```
LukeTowers: What was the reason for this whitespace change?

multiwebinc: Likely just a mistake.

mjauvin:
@multiwebinc I am slowly making progress at adding a column to show the "found" db field...

mjauvin:
@multiwebinc I'll let you play with this, but so far I did the following:

In  `controllers/messages/config_table.yaml` modify columns as below:
```
columns:
    from:
        title: From
    to: 
        title: To
    found:
        title: Found
        type: checkbox
        width: 100px
```

And in `controllers/Messages.php` add the found column data in `processTableData()` as below:
```
$data[] = [
    'id' => $message->id,
    'code' => $message->code,
    'from' => $message->forLocale($fromCode),
    'to' => $toContent, 
    'found' => $message->found
];
```

checkbox column type should be able to be marked "readOnly" but it's not working.

It's currently rendering as below:

Screenshot from 2020-03-27 18-29-28

multiwebinc:
@mjauvin I couldn't find any [documentation](https://octobercms.com/docs/backend/widgets) for the [backend table widget](https://octobercms.com/docs/api/backend/widgets/table), however the supported column types are "string", "checkbox", "dropdown" and "autocomplete". These are all rendered client-side with JS and the data is pulled using AJAX. If the "string" column type is chosen, it is HTML entity encoded, so it doesn't look like there's a straightforward and easy way to do what I was thinking. Using a read-only checkbox would be my second choice, however the read-only attribute has no effect for the checkbox column type. In fact, selecting/unselecting the checkbox does not update the value in the database either. Looks like a possible bug to me.

multiwebinc:
Ok, this is how it is rendering now. I put the info icon with a `title` attribute explaining what this means in the table header:

image

mjauvin:
@multiwebinc I finally had time to test this and it works well.

I think the column header would make more sense being "notes" rather than "found" since the later sounds more like it should say it was found (yes) or not (no)... only marking the records that have not been found is confusing when using that column header in my opinion (especially if there are no missing translations).

Other that that, it looks great!

controllers/messages/config_table.yaml
```diff
@@ -13,3 +13,6 @@ columns:
         title: From
     to:
         title: To
+    found:
+        title: Found
```
mjauvin, line 17: We should use a translation string here i.e. rainlab.translate::lang.messages.found_title

multiwebinc:
@mjauvin The thing is that if it wasn't found with the scanner, I would like to be able to provide a slightly more descriptive feedback to the user than just "Not found" since I would guess the majority of users wouldn't know what that means exactly. What if we call it "Scan errors", or "Scan result"?

mjauvin:
> @mjauvin The thing is that if it wasn't found with the scanner, I would like to be able to provide a slightly more descriptive feedback to the user than just "Not found" since I would guess the majority of users wouldn't know what that means exactly. What if we call it "Scan errors", or "Scan result"?

Whatever makes sense, I was just saying "Found" didn't make sense here.

controllers/messages/config_table.yaml
multiwebinc: Translations are not necessary because the actual title comes from [this line](https://github.com/multiwebinc/translate-plugin/blob/7fc5503dc78ab8045ae202b378982d7d288920d7/controllers/messages/_table_headers.htm#L104). I am assuming they are only used here as a fallback in case something down the line should fail for whatever reason. This line can actually be commented out and it will still work.

multiwebinc:
@mjauvin Take a look now. Let me know if the wording sounds good to you now.

mjauvin:
Looks good @multiwebinc!

PR #566, opened 2020-03-27
Generate message code using md5()

### LINKED ISSUE #563: Changing non-alphanumeric characters does not change the string

mjauvin:
- generate md5 hash of message id for message code
- includes migration file

fixes #563
replaces #564

Potential breaking change with Message Ids. Message Ids are now encoded using `md5()` to generate the Message code. Message Ids that contain characters other than letters, numbers, white spaces and `.` might have previously been sharing a Message translation due to their resulting Message code being equal. These sharing are no longer going to happen. Furthermore, if you were querying the Message model directly by code, those queries will now fail since the codes are now 32 hexadecimal characters (md5 hash of the Message Id).


multiwebinc:
Wouldn't it be better to move `Message::makeMessageCode()` to the migration since that's the only place it should be needed now and should not need to be maintained within the actual Message class?

multiwebinc:
`Str` is no longer needed in the `Message` model.

multiwebinc:
Also, the test `testMakeMessageCode` should probably be removed.

mjauvin:
> Also, the test `testMakeMessageCode` should probably be removed.

That was done in the last commit

tests/unit/models/MessageTest.php
```diff
@@ -1,61 +0,0 @@
-<?php namespace RainLab\Translate\Tests\Unit\Models;
```
LukeTowers: The test should be rewritten not removed. Has @daftspunk provided input on this change?

mjauvin: No input on the change yet. I'll adapt the tests

models/Message.php
```diff
@@ -276,23 +275,12 @@ protected static function makeCacheKey()
...
-        $messageId = preg_replace('!['.preg_quote('_').'|'.preg_quote('-').']+!u', $separator, $messageId);
-
-        // Remove all characters that are not the separator, letters, numbers, or whitespace.
-        $messageId = preg_replace('![^'.preg_quote($separator).'\pL\pN\s]+!u', '', mb_strtolower($messageId));
-
-        // Replace all separator characters and whitespace by a single separator
-        $messageId = preg_replace('!['.preg_quote($separator).'\s]+!u', $separator, $messageId);
-
-        return Str::limit(trim($messageId, $separator), 250);
+        return md5(trim($messageId));
```
mjauvin: I wonder if the generated code should be case insensitive (as previous behavior)?
i.e. `return md5(trim(str_tolower($messageId)));`

multiwebinc: What would the advantage of case-insensitivity be? And could there be a use case for case-sensitivity? For example, if the original author writes a string in all caps and then later decides that it looks silly that way, so they update the string. I would think that the translators should be aware of that.

multiwebinc: Well, the same could be said for punctuation as well. People might have "My string!" in one place and "My string" in another place. I think that if you're going to create a change that could potentially break things, you might as well go all in so that the breaks occur on a single update. It's probably either now or never if you're going to change this.

mjauvin: True, but I feel there is more potential for case sensitive ids than ones with punctuation.

In any case, we'll need @daftspunk's input on this breaking change.

multiwebinc: @mjauvin Just a thought. What about adding a setting somewhere to set whether the messages are case-sensitive or not? Then the default could be set to insensitive and allow the user to set it as sensitive should the case arise.

mjauvin:
@daftspunk do you see a problem merging this?

mjauvin:
Getting ready to merge this for release 1.8.0, any objections?

I've tested it pretty well, and unit tests have been added as well.

LukeTowers:
Isn't this a breaking change?

mjauvin:
@LukeTowers It's not, unless multiple different messageIds resulted in the same code after characters other than  letters, numbers, white spaces and "." got removed in the old `makeMessageCode()` method.

For example, using the old `makeMessageCode` method, the following messageIds would have resulted in the same code:

- "my-message-key"
- "my=message=key"
- "my message key"
- "my.message.key"

This is no longer the case since we use md5() on the messageId.

mjauvin:
@bennothommo any idea what's happening with the tests in here suddenly?

LukeTowers:
The breaking change is in the changes made in the migration is it not? If someone is using `Message::whereCode()` manually, wouldn't that break as soon as the migration runs? @bennothommo what do you think, am I being too paranoid? Also does this affect the codes presented in the theme string translations section?

bennothommo:
@LukeTowers it is a breaking change, but I think @mjauvin has appropriately updated it through making this an update that requires confirmation, so as long as people are made clearly aware of the change, we should be alright.

However - unless there's something I'm not considering - couldn't `whereCode` be adjusted to be compatible with previous usage scenarios? TBH, the Translate plugin is the plugin I use the least (ie. pretty much not at all) because I don't (and neither does my company) do multi-lang websites, so I'm not fully across it all.

mjauvin:
@LukeTowers @bennothommo where do you see this `whereCode()` method?

@LukeTowers the theme string translations section uses messageIds (like in twig or in php code) and those get converted into a code by `makeMessageCode() `

mjauvin:
@bennothommo @LukeTowers I don't understand why the tests are now failing, it makes no sense. What am I missing?

updates/version.yaml
```diff
@@ -100,3 +100,7 @@
     - Polish localization updates
     - Turkish localization updates
     - Add Greek language localization
+1.8.0:
+    - "!!! Potential breaking change with Message Ids. Message Ids are now encoded using `md5()` to generate the Message code. Message Ids that contain characters other than letters, numbers, white spaces and `.` might have previously been sharing a Message translation due to their resulting Message code being equal. These sharing are no longer going to happen. Furthermore, if you were querying the Message model directly by code, those queries will now fail since the codes are now 32 hexadecimal characters (md5 hash of the Message Id)."
```
LukeTowers, line 104:
```suggestion
    - "!!! Potential breaking change: Message codes are now MD5 hashed versions of the original string. See github.com/rainlab/translate-plugin/pull/556 for details."
```

models/Message.php
```diff
@@ -289,23 +288,16 @@ protected static function makeCacheKey()
...
-
-        // Replace all separator characters and whitespace by a single separator
-        $messageId = preg_replace('!['.preg_quote($separator).'\s]+!u', $separator, $messageId);
-
-        return Str::limit(trim($messageId, $separator), 250);
+        if (Config::get('rainlab.translate::caseSensitiveIds', false)) {
+            return md5(trim($messageId));
+        } else {
+            return md5(mb_strtolower(trim($messageId)));
+        }
```
LukeTowers, line 301: Why was this change required? Also wasn't the previous behavior to be case sensitive? Wouldn't you want these codes to be case sensitive?

mjauvin: @LukeTowers previous behavior was case INSENSITIVE and is preserved by default. See old code below:
```
        $messageId = preg_replace('![^'.preg_quote($separator).'\pL\pN\s]+!u', '', mb_strtolower($messageId));
```

mjauvin: The change is not REQUIRED but gives the option to developpers to change the behavior.

LukeTowers: Why would we want to provide the option when we haven't up to this point? Personally I'd think that case sensitive should have been the default from the start, but now that it isn't, what's the benefit of allowing it to be?

mjauvin: You kind of just answered that. If you think this should have been the default, why not give that option now? It doesn't hurt anything.

LukeTowers: I'd be concerned about the message ID changing whenever the config item value changes

mjauvin:
Good point. It shouldn't be changed once a project has started, but nothing can prevent it and it would cause mayhem... I'll revert this change.

mjauvin:
The reason I was having trouble with the migration file during tests is that on the `down()` method, the table didn't exist anymore (not sure exactly why, but it's not the first time I see this behavior under the github test environment).

tests/unit/models/MessageTest.php
```diff
@@ -11,51 +11,32 @@ public function testImportMessages()
...
-        $this->assertEquals('щука', Message::makeMessageCode('щука'));
-        $this->assertEquals('foo.漢字', Message::makeMessageCode('foo 漢字')); // Chinese
-        $this->assertEquals('xin.chào.thế.giới', Message::makeMessageCode('xin chào thế giới'));
-        $this->assertEquals('xin.chào.thế.giới', Message::makeMessageCode('XIN CHÀO THẾ GIỚI'));
-        $this->assertEquals('đấm.phát.chết.luôn', Message::makeMessageCode('đấm phát chết luôn'));
-        $this->assertEquals('foo', Message::makeMessageCode('foo ')); // no-break space (U+00A0)
-        $this->assertEquals('foo', Message::makeMessageCode('foo           ')); // spaces U+2000 to U+200A
-        $this->assertEquals('foo', Message::makeMessageCode('foo ')); // narrow no-break space (U+202F)
-        $this->assertEquals('foo', Message::makeMessageCode('foo ')); // medium mathematical space (U+205F)
-        $this->assertEquals('foo', Message::makeMessageCode('foo　')); // ideographic space (U+3000)
```
LukeTowers, line 59: Why remove all of these cases?

updates/migrate_message_code.php
```diff
@@ -0,0 +1,54 @@
...
+
+class MigrateMessageCode extends Migration
+{
+    const TABLE_NAME = 'rainlab_translate_messages';
+
+    public function up()
+    {
+        if (!Schema::hasTable(self::TABLE_NAME)) {
+            return;
+        }
```
LukeTowers, line 16: Why would this be required in the up() method?

updates/migrate_message_code.php
```diff
@@ -0,0 +1,54 @@
...
+
+        // Remove all characters that are not the separator, letters, numbers, or whitespace.
+        $messageId = preg_replace('![^'.preg_quote($separator).'\pL\pN\s]+!u', '', mb_strtolower($messageId));
+
+        // Replace all separator characters and whitespace by a single separator
+        $messageId = preg_replace('!['.preg_quote($separator).'\s]+!u', $separator, $messageId);
+
+        return Str::limit(trim($messageId, $separator), 250);
+    }
+
```
LukeTowers, line 53:
```suggestion
```

models/Message.php
```diff
@@ -289,23 +288,12 @@ protected static function makeCacheKey()
     }
 
     /**
-     * Creates a sterile key for a message.
+     * Creates an md5 representation of the message Id
```
LukeTowers, line 291:
```suggestion
     * Get the code for the provided message (32 character hexadecimal MD5-hashed version)
```

models/Message.php
```diff
@@ -289,23 +288,12 @@ protected static function makeCacheKey()
...
      * @param  string $messageId
```
LukeTowers, line 292:
```suggestion
     * @param  string $message
```

models/Message.php
```diff
@@ -289,23 +288,12 @@ protected static function makeCacheKey()
...
      * @return string
      */
-    protected static function makeMessageCode($messageId)
+    public static function makeMessageCode($messageId)
```
LukeTowers, line 295:
```suggestion
    public static function makeMessageCode($message)
```

models/Message.php
```diff
@@ -289,23 +288,12 @@ protected static function makeCacheKey()
...
-        $messageId = preg_replace('!['.preg_quote('_').'|'.preg_quote('-').']+!u', $separator, $messageId);
-
-        // Remove all characters that are not the separator, letters, numbers, or whitespace.
-        $messageId = preg_replace('![^'.preg_quote($separator).'\pL\pN\s]+!u', '', mb_strtolower($messageId));
-
-        // Replace all separator characters and whitespace by a single separator
-        $messageId = preg_replace('!['.preg_quote($separator).'\s]+!u', $separator, $messageId);
-
-        return Str::limit(trim($messageId, $separator), 250);
+        return md5(mb_strtolower(trim($messageId)));
```
LukeTowers, line 297:
```suggestion
        return md5(mb_strtolower(trim($message)));
```

tests/unit/models/MessageTest.php
mjauvin: Because they are no longer relevant.

mjauvin: The tests were there to make sure the previous encoding was doing what it was supposed to.

LukeTowers: Would they still pass so long as the MD5 versions were used for comparison?

mjauvin: They would, but what's the point of comparing the md5 hash of a string with the result of a method which does pretty much only return the md5 hash of its argument?

PR #569, opened 2020-04-08
Fixed translation message table title identifier

Eoler:


mjauvin:
@multiwebinc is that something you missed in your last PR?

mjauvin:
Thanks @Eoler

multiwebinc:
@mjauvin Most likely, yes. I changed the string name from `found` to `found_title` because you recommended changing the actual string from "Found" to something else.

PR #572, opened 2020-04-16
Fix docs: translate key missing quotes

drmzio:
Just spent a good hour trying to figure out why the translate key wasn't outputting the translated text. The translate key must be in quotes when calling it inside the translate twig filter. 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ If translated, the text above will appear as whatever language is selected by th
 
 Message or string translation is the conversion of adhoc strings used throughout the site. A message can be translated with parameters.
 
-    {{ site.name|_ }}
+    {{ 'site.name'|_ }}
 
     {{ 'Welcome to our website!'|_ }}
 


mjauvin:
site.name is a variable here, it should not be quoted. The next example shows a string key.

drmzio:
@mjauvin Are you sure about that? Just checked `site` and `site.name` and they're never declared as variables anywhere. The example provided in the docs reference the `site.name` key in the `theme.yaml`

```yaml
translate:
    en:
        site.name: 'My Website'
        nav.home: 'Home'
        nav.video: 'Video'
        title.home: 'Welcome Home'
        title.video: 'Screencast Video'
```

The only way to pull the values is by calling `{{ 'site.name'|_ }}` in quotes.

mjauvin:
If it is referring to that key in the theme translation, then you are correct. My appology.

PR #576, opened 2020-04-21
Fired events in the theme scanner.

bereznev:
Fired events in ThemeScanner.php class for ability to  extend scan method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -275,7 +275,8 @@ It's possible to translate query string parameters by listening to the `translat
 For a possible implementation of the `YourModel::translateParams` method look at the example under `URL parameter translation` from above.
 
 ## Extend theme scan
-      Event::listen('rainlab.translate.afterScan', function (ThemeScanner $scanner) {
+
+      Event::listen('rainlab.translate.themeScanner.afterScan', function (ThemeScanner $scanner) {
            ...
       });
 
--- a/README.md
+++ b/README.md
@@ -274,6 +274,11 @@ It's possible to translate query string parameters by listening to the `translat
 
 For a possible implementation of the `YourModel::translateParams` method look at the example under `URL parameter translation` from above.
 
+## Extend theme scan
+      Event::listen('rainlab.translate.afterScan', function (ThemeScanner $scanner) {
+           ...
+      });
+
 ## Settings model translation
 
 It's possible to translate your settings model like any other model. To retrieve translated values use:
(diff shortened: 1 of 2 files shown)


LukeTowers:
@bereznev please provide a real world use case for both proposed events. Additionally, please add inline event documentation (example here: https://github.com/octobercms/october/blob/develop/modules/backend/classes/Controller.php#L245-L263)

bereznev:
I added an extra scan for the php classes. Using it allows you to add words list using constants (properties) to prevent duplication of these lists. And I need to automatically generate translation files after scanning.

LukeTowers:
Please provide the code you use to listen to the event so I can better understand the purpose.

bereznev:
`        Event::listen('translate.themeScanner.afterScan', function (ThemeScanner $scanner) {
            $dir = themes_path('my-theme/assets/lang');
            if (!is_dir($dir)) {
                File::makeDirectory($dir);
            }
            $exportMode = new MessageExport();
            $result = $exportMode->exportData(['code', 'sk', 'en', 'x',]);

            foreach (['sk', 'en'] as $lang) {
                $content = collect($result)->keyBy('x')->map(function (array $item) use ($lang) {
                    return $item[$lang];
                })->toArray();
                File::put("{$dir}/{$lang}.json", json_encode($content));
            }
        });`

classes/ThemeScanner.php
```diff
@@ -24,7 +25,20 @@ public static function scan()
...
+         * Fires after theme scanning.
+         *
+         * Example usage:
+         *
+         *     Event::listen('translate.themeScanner.afterScan', function (ThemeScanner $scanner) {
+         *         // added an extra scan. Add generation files...
+         *     });
+         *
+         */
+        Event::fire('translate.themeScanner.afterScan', [$obj]);
```
LukeTowers, line 41: Event should be `rainlab.translate.themeScanner.afterScan`. Also you need to continue to return the result of the scan method

bennothommo: @LukeTowers I was going to mention the return thing too, but the method returns `void`, so there was no need for a `return` call in the first place.

LukeTowers:
Change the event names in all the places it's referenced please.

README.md
```diff
@@ -274,6 +274,11 @@ It's possible to translate query string parameters by listening to the `translat
 
 For a possible implementation of the `YourModel::translateParams` method look at the example under `URL parameter translation` from above.
 
+## Extend theme scan
+      Event::listen('rainlab.translate.afterScan', function (ThemeScanner $scanner) {
```
LukeTowers, line 278: Event name still needs to be changed here. Please also add a newline between the header and the code example.

LukeTowers:
LGTM

PR #577, opened 2020-04-28
Allows to export default locale

Samuell1:
Fixes #422 

Reworked original PR from few years ago #431


mjauvin:
Samuell, it would need to be triggered by an additional checkbox to be accepted, otherwise it will break the current behavior. (e.g. include original key)

Samuell1:
@mjauvin yes you can disable it or enabled in export menu.

mjauvin:
@Samuell1 could you set this new column to UNCHECKED by default to keep existing behavior?

Samuell1:
@mjauvin Not sure if ocms supports that for export.

PR #587, opened 2020-07-05
Fix transOrderBy could return duplicates for models with multiple indexes

aurelien-roy:
The condition on `rainlab_translate_indexes.item` (present in `transWhere`) was missing in `transOrderBy`. This could lead to `Model::transOrderBy('field')->get()` returning duplicates if model has more than one index.

This commit moves the condition on `rainlab_translate_indexes.item` into the `joinTranslateIndexesTable` method so that it gets applied on both `transOrderBy` and `transWhere`.


mjauvin:
Merged into #586

PR #605, opened 2020-09-21
mail templates/views can now be localized

### LINKED ISSUE #533: Mail Templates translation

mjauvin:
fixes #533

### DIFF
--- a/Plugin.php
+++ b/Plugin.php
@@ -37,6 +37,13 @@ class Plugin extends PluginBase
 
     public function register()
     {
+        /* 
+         * Load localize version of mail templates (akin to localized CMS content files)
+         */
+        Event::listen('mailer.beforeAddContent', function ($mailer, $message, $view, $data, $raw, $plain) {
+            return EventRegistry::instance()->addLocalizedContent($mailer, $message, $view, $data, $raw, $plain);
+        }, 1);
+
         /*
          * Defer event with low priority to let others contribute before this registers.
          */
--- a/Plugin.php
+++ b/Plugin.php
@@ -38,7 +38,7 @@ class Plugin extends PluginBase
     public function register()
     {
         /* 
-         * Load localize version of mail templates (akin to localized CMS content files)
+         * Load localized version of mail templates (akin to localized CMS content files)
          */
         Event::listen('mailer.beforeAddContent', function ($mailer, $message, $view, $data, $raw, $plain) {
             return EventRegistry::instance()->findLocalizedMailViewContent($mailer, $message, $view, $data, $raw, $plain);
--- a/Plugin.php
+++ b/Plugin.php
@@ -41,7 +41,7 @@ class Plugin extends PluginBase
          * Load localize version of mail templates (akin to localized CMS content files)
          */
         Event::listen('mailer.beforeAddContent', function ($mailer, $message, $view, $data, $raw, $plain) {
-            return EventRegistry::instance()->addLocalizedContent($mailer, $message, $view, $data, $raw, $plain);
+            return EventRegistry::instance()->findLocalizedMailViewContent($mailer, $message, $view, $data, $raw, $plain);
         }, 1);
 
         /*
--- a/classes/EventRegistry.php
+++ b/classes/EventRegistry.php
@@ -1,15 +1,19 @@
 <?php namespace RainLab\Translate\Classes;
 
-use Str;
+use App;
+use Exception;
 use File;
+use Mail;
+use Request;
+use Str;
 use Cms\Classes\Page;
 use Cms\Classes\Content;
+use System\Classes\MailManager;
 use System\Classes\PluginManager;
 use RainLab\Translate\Models\Message;
 use RainLab\Translate\Models\Locale as LocaleModel;
 use RainLab\Translate\Classes\Translator;
 use RainLab\Translate\Classes\ThemeScanner;
-use Exception;
 
 /**
  * Registrant class for bootstrapping events
@@ -262,5 +266,35 @@ class EventRegistry
         return $templates->filter(function($template) use ($extensions) {
             return !Str::endsWith($template->getBaseFileName(), $extensions);
         });
+    }
+
+    public function addLocalizedContent($mailer, $message, $view, $data, $raw, $plain)
+    {
+        if ($raw !== null || $view === null || !is_string($view)) {
+            return;
+        }
+
+        if (App::runningInBackend()) {
+            list($defaultLocale) = explode('-', Request::getDefaultLocale());
+            list($locale) = explode('-', App::getLocale());
+        } else {
+            $translator = Translator::instance();
+            $locale = $translator->getLocale();
+            $defaultLocale = $translator->getDefaultLocale();
+        }
+
+        if ($locale === $defaultLocale) {
+            return;
+        }
+
+        $view = sprintf('%s-%s', $view, $locale);
+
+        $factory = Mail::getViewFactory();
+        if (!$factory->exists($view)) {
+            return;
+        }
+
+        $mailManager = MailManager::instance();
+        return !$mailManager->addContentToMailer($message, $view, $data, false);
     }
 }
--- a/classes/EventRegistry.php
+++ b/classes/EventRegistry.php
@@ -302,9 +302,8 @@ class EventRegistry
 
         $plainOnly = empty($view);
 
-        $result = MailManager::instance()->addContentToMailer($message, $code, $data, $plainOnly);
-        if ($result) {
-            // prevent the caller who fired the event from continuing to send the mail.
+        if (MailManager::instance()->addContentToMailer($message, $code, $data, $plainOnly)) {
+            // the caller who fired the event is expecting a FALSE response to halt the event
             return false;
         }
     }
--- a/classes/EventRegistry.php
+++ b/classes/EventRegistry.php
@@ -282,7 +282,7 @@ class EventRegistry
             return null;
         }
 
-        $locale = strtolower(isset($data['_current_locale']) ? $data['_current_locale'] : App::getLocale());
+        $locale = !empty($data['_current_locale']) ? $data['_current_locale'] : App::getLocale();
 
         $factory = $mailer->getViewFactory();
 
@@ -312,6 +312,8 @@ class EventRegistry
      */
     public function getLocalizedView($factory, $code, $locale)
     {
+        $locale = strtolower($locale);
+
         $searchPaths[] = $locale;
 
         if (str_contains($locale, '-')) {
--- a/classes/EventRegistry.php
+++ b/classes/EventRegistry.php
@@ -3,8 +3,6 @@
 use App;
 use Exception;
 use File;
-use Mail;
-use Request;
 use Str;
 use Cms\Classes\Page;
 use Cms\Classes\Content;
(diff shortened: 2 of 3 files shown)


mjauvin:
Blocked by https://github.com/octobercms/library/pull/523

classes/EventRegistry.php
```diff
@@ -263,4 +265,67 @@ public function pruneTranslatedContentTemplates($templates)
...
+            $view = $this->getLocalizedView($factory, $view, $locale);
+        }
+
+        if (isset($plain)) {
+            $plain = $this->getLocalizedView($factory, $plain, $locale);
+        }
+
+        if (($code = $view ?: $plain) === null) {
+            return;
+        }
```
LukeTowers, line 299: What case are you trying to handle here?

classes/EventRegistry.php
```diff
@@ -263,4 +265,67 @@ public function pruneTranslatedContentTemplates($templates)
...
+
+        return !MailManager::instance()->addContentToMailer($message, $code, $data, $view === null);
```
LukeTowers, line 301: What parameter is being based by $view === null? And why is the return result negated?

classes/EventRegistry.php
```diff
@@ -263,4 +265,67 @@ public function pruneTranslatedContentTemplates($templates)
...
+        }
+
+        foreach ($searchPaths as $path) {
+            $localizedView = sprintf('%s-%s', $code, $path);
+
+            if ($factory->exists($localizedView)) {
+                return $localizedView;
+            }
+        }
+        return;
```
LukeTowers, line 329:
```suggestion
        return null;
```

LukeTowers (changes requested):

classes/EventRegistry.php
```diff
@@ -263,4 +265,67 @@ public function pruneTranslatedContentTemplates($templates)
...
+     * @param  string $plain
+     * @return string|null
+     */
+    public function findLocalizedMailViewContent($mailer, $message, $view, $data, $raw, $plain)
+    {
+        if (isset($raw) || (!isset($view) && !isset($plain))) {
+            return;
+        }
+
+        $locale = strtolower(isset($data['_current_locale']) ? $data['_current_locale'] : App::getLocale());
```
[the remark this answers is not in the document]
mjauvin: I need the result to be lowercase...

classes/EventRegistry.php
```diff
@@ -263,4 +265,67 @@ public function pruneTranslatedContentTemplates($templates)
...
+     * @param  array $data
+     * @param  string $raw
+     * @param  string $plain
+     * @return string|null
+     */
+    public function findLocalizedMailViewContent($mailer, $message, $view, $data, $raw, $plain)
+    {
+        if (isset($raw) || (!isset($view) && !isset($plain))) {
+            return;
+        }
```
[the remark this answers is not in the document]
mjauvin: from browsing the source code when developping this, I used those conditions because it can happen.

classes/EventRegistry.php
mjauvin: $plainOnly
```
modules/system/classes/MailManager.php:    protected function addContentToMailerInternal($message, $template, $data, $plainOnly = false)
```
the return is negated because when `addContentToMailer()` returns true, it means the content has been successfully added and we don't want the function firing the event continue its execution (see below)

src/Mail/Mailer.php:
```
        if (
            ($this->fireEvent('mailer.beforeAddContent', [$message, $view, $data, $raw, $plain], true) === false) ||
            (Event::fire('mailer.beforeAddContent', [$this, $message, $view, $data, $raw, $plain], true) === false)
        ) { 
            return;
        }
```

classes/EventRegistry.php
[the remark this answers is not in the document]
mjauvin: you may be right, I simplified this.

classes/EventRegistry.php
mjauvin: if no localized html view or plain view is found, we return so the regular view is rendered.

classes/EventRegistry.php
LukeTowers: Why return anything then if we don't want the execution to stop after the event handler is processed? addContentToMailer can return false, which means that this handler would stop execution if it failed to do its job?

mjauvin: if addContentToMailer() returns true, then it sent the mail using the localized view.

if addContentToMailer() returns false, then we WANT the regular flow to continue to send the mail using the original "non-localized" view.

mjauvin:
Why did you add the blocked status, the PR for which it was blocked has been merged...

classes/EventRegistry.php
LukeTowers: If that's the case then why is there the `!` in front of it?

mjauvin: I just explained it above @LukeTowers . See my last commit, it makes it more readable.

LukeTowers:
@mjauvin it's blocked pending the completion of our discussion ;) I suppose review needed is probably a better status tag to apply though

classes/EventRegistry.php
```diff
@@ -263,4 +265,77 @@ public function pruneTranslatedContentTemplates($templates)
...
+            return null;
+        }
+
+        $plainOnly = empty($view);
+
+        $result = MailManager::instance()->addContentToMailer($message, $code, $data, $plainOnly);
+        if (!$result) {
+            // prevent the caller who fired the event from continuing to send the mail.
+            return false;
+        }
```
LukeTowers, line 309: This has only confused me more. Now you're saying that if the call returns false, you want to return false; but you're also saying return nothing unless the call returns falsey. Either way the return value of this method will always be falsey and thus never halt the event. Are you sure you don't just mean return the result of `MailManager::instance()->addContentToMailer()` unaltered?

mjauvin: But Luke, halting the event IS done by returning false... what  am I missing here?

mjauvin: but in that last commit I did reverse the logic which was fine before... 🤣

mjauvin: should be good now.

LukeTowers: > But Luke, halting the event IS done by returning false... what am I missing here?

The question is what was I missing here 😂. I didn't read the entire line of code for the firing and completely missed the `=== false`

classes/EventRegistry.php
```diff
@@ -263,4 +265,77 @@ public function pruneTranslatedContentTemplates($templates)
...
+     * @return string|null
+     */
+    public function findLocalizedMailViewContent($mailer, $message, $view, $data, $raw, $plain)
+    {
+        // There is no need to localize raw templates
+        if (!empty($raw)) {
+            return null;
+        }
+
+        $locale = !empty($data['_current_locale']) ? $data['_current_locale'] : App::getLocale();
```
LukeTowers, line 286:
```suggestion
        // Get the locale to use for this template
        $locale = !empty($data['_current_locale']) ? $data['_current_locale'] : App::getLocale();
```

classes/EventRegistry.php
```diff
@@ -263,4 +265,77 @@ public function pruneTranslatedContentTemplates($templates)
...
+     * @param  \Illuminate\Mail\Message $message
+     * @param  string $view
+     * @param  array $data
+     * @param  string $raw
+     * @param  string $plain
+     * @return string|null
+     */
+    public function findLocalizedMailViewContent($mailer, $message, $view, $data, $raw, $plain)
+    {
+        // There is no need to localize raw templates
```
LukeTowers, line 281:
```suggestion
        // Raw content cannot be localized at this level
```

classes/EventRegistry.php
```diff
@@ -263,4 +265,77 @@ public function pruneTranslatedContentTemplates($templates)
...
+        if (!empty($raw)) {
+            return null;
```
LukeTowers, line 283:
```suggestion
            return;
```

classes/EventRegistry.php
```diff
@@ -263,4 +265,77 @@ public function pruneTranslatedContentTemplates($templates)
...
+
+    /**
+     * Adds language suffixes to mail view files.
+     * @param  \October\Rain\Mail\Mailer $mailer
+     * @param  \Illuminate\Mail\Message $message
+     * @param  string $view
+     * @param  array $data
+     * @param  string $raw
+     * @param  string $plain
+     * @return string|null
```
LukeTowers, line 277:
```suggestion
     * @return bool|void Will return false if the translation process successfully replaced the original message with a translated version to prevent the original version from being processed.
```

classes/EventRegistry.php
```diff
@@ -263,4 +265,77 @@ public function pruneTranslatedContentTemplates($templates)
...
+        // There is no need to localize raw templates
+        if (!empty($raw)) {
+            return null;
+        }
+
+        $locale = !empty($data['_current_locale']) ? $data['_current_locale'] : App::getLocale();
+
+        $factory = $mailer->getViewFactory();
+
+        if (isset($view)) {
```
LukeTowers, line 290:
```suggestion
        if (!empty($view)) {
```

classes/EventRegistry.php
```diff
@@ -263,4 +265,77 @@ public function pruneTranslatedContentTemplates($templates)
...
+            $view = $this->getLocalizedView($factory, $view, $locale);
+        }
+
+        if (isset($plain)) {
```
LukeTowers, line 294:
```suggestion
        if (!empty($plain)) {
```

classes/EventRegistry.php
mjauvin: Ok, I was getting worried... :relieved:

classes/EventRegistry.php
```diff
@@ -263,4 +265,77 @@ public function pruneTranslatedContentTemplates($templates)
...
+            return null;
+        }
+
+        $plainOnly = empty($view);
+
+        $result = MailManager::instance()->addContentToMailer($message, $code, $data, $plainOnly);
+        if ($result) {
+            // prevent the caller who fired the event from continuing to send the mail.
+            return false;
+        }
```
LukeTowers, line 309:
```suggestion
        $code = $view ?: $plain;
        if (empty($code)) {
            return null;
        }

        $plainOnly = empty($view);

        // If the content swap process succeeded, return false prevent the original content from being processed
        return !MailManager::instance()->addContentToMailer($message, $code, $data, $plainOnly);
```

LukeTowers (changes requested):

LukeTowers:
LGTM now, @bennothommo any thoughts?

LukeTowers (approved):

Plugin.php
```diff
@@ -37,6 +37,13 @@ public function pluginDetails()
 
     public function register()
     {
+        /* 
+         * Load localize version of mail templates (akin to localized CMS content files)
```
bennothommo, line 41:
```suggestion
         * Load localized version of mail templates (akin to localized CMS content files)
```

bennothommo (approved):

bennothommo:
Looks good to me, guys. Well done @mjauvin.

PR #607, opened 2020-09-26
simplify code to replace ml field types

mjauvin:


classes/EventRegistry.php
```diff
@@ -165,24 +174,7 @@ protected function processFormMLFields($fields, $model)
...
-            elseif ($type == 'markdown') {
-                $fields[$name]['type'] = 'mlmarkdowneditor';
-            }
-            elseif ($type == 'repeater') {
-                $fields[$name]['type'] = 'mlrepeater';
-            }
-            elseif ($type == 'mediafinder') {
-                $fields[$name]['type'] = 'mlmediafinder';
-            }
+            $fields[$name]['type'] = $typesMap[$type];
```
LukeTowers, line 177: You'd have to wrap this in a conditional to check to see if there is a mapped type for the current type before replacing it.

PR #608, opened 2020-10-02
add missing allowedExtensions in MLContent

mjauvin:
fixes https://github.com/rainlab/pages-plugin/issues/453

update migration files with incorrect syntax

### DIFF
--- a/tests/unit/behaviors/TranslatablePageTest.php
+++ b/tests/unit/behaviors/TranslatablePageTest.php
@@ -22,13 +22,6 @@ class TranslatablePageTest extends PluginTestCase
         $resolver = new Resolver(['theme1' => $datasource]);
         $resolver->setDefaultDatasource('theme1');
         Model::setDatasourceResolver($resolver);
-
-        TranslatablePage::extend(function($page) {
-            if (!$page->isClassExtendedWith('RainLab\Translate\Behaviors\TranslatablePage')) {
-                $page->addDynamicProperty('translatable', ['title']);
-                $page->extendClassWith('RainLab\Translate\Behaviors\TranslatablePage');
-            }
-        });
     }
 
     public function tearDown(): void
--- a/updates/update_messages_table.php
+++ b/updates/update_messages_table.php
@@ -5,19 +5,30 @@ use October\Rain\Database\Updates\Migration;
 
 class UpdateMessagesTable extends Migration
 {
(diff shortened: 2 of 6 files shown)


tests/unit/behaviors/TranslatablePageTest.php
```diff
@@ -22,13 +22,6 @@ public function setUp(): void
         $resolver = new Resolver(['theme1' => $datasource]);
         $resolver->setDefaultDatasource('theme1');
         Model::setDatasourceResolver($resolver);
-
-        TranslatablePage::extend(function($page) {
-            if (!$page->isClassExtendedWith('RainLab\Translate\Behaviors\TranslatablePage')) {
-                $page->addDynamicProperty('translatable', ['title']);
-                $page->extendClassWith('RainLab\Translate\Behaviors\TranslatablePage');
-            }
```
LukeTowers, line 30: What was the reason for this change @mjauvin?

mjauvin:
@LukeTowers This has been handled in Plugin.php for a while now.
```
        /*
         * Handle translated page URLs
         */
        Page::extend(function($page) {
            if (!$page->propertyExists('translatable')) {
                $page->addDynamicProperty('translatable', []);
            }
            $page->translatable = array_merge($page->translatable, ['title', 'description', 'meta_title', 'meta_description']);
            $page->extendClassWith('RainLab\Translate\Behaviors\TranslatablePageUrl');
            $page->extendClassWith('RainLab\Translate\Behaviors\TranslatablePage');
        });
```

This was causing the test to fail because the Page class had already been extended with the TranslatablePage behavior.

updates/update_messages_table.php
```diff
@@ -5,19 +5,30 @@
...
     {
-        Schema::table('rainlab_translate_messages', function($table)
-        {
-            $table->dropColumn('found');
-        });
+        if (!Schema::hasTable(self::TABLE_NAME)) {
+            return;
+        }
+
+        if (Schema::hasColumn(self::TABLE_NAME, 'found')) {
```
LukeTowers, line 30: @mjauvin just as an FYI, I know it doesn't really matter since migration files should probably never be extended, but I would prefer we always use `static::` over `self::` unless there's a really good reason to use `self::`. It's a better practice to keep since that will provide the most flexibility for developers looking to extend the code.

PR #619, opened 2020-10-30
Update lang/pl/lang.php

mikolajklek:
Fixed typos, and uptaded pl language to latest version

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -45,7 +45,7 @@ jobs:
         with:
           php-version: ${{ matrix.phpVersions }}
           extensions: ${{ env.phpExtensions }}
-          tools: composer
+          tools: composer:v1
           coverage: none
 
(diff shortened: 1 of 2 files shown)


mjauvin:
Please leave the opening php tag on its own line and keep the spacing before the return.

LukeTowers:
@mjauvin I will note that I personally prefer that the return statement is on the same line as the opening PHP tag for language files, but we haven't enforced one way or the other throughout our codebase yet.

LukeTowers:
@mjauvin if you're up for it these tests need to be fixed by forcing Composer to 1.0 for now like Ben did in the main repo, or you could test the Composer 2.0 fix PR in the wikimedia-merge-plugin with October so that we can move onto the next version of that plugin and the tests will fix themselves.

mjauvin:
Thanks for your contribution @mikolajklek!

PR #623, opened 2020-11-13
Faster version of transWhere

acasar:
This PR is an alternative version of https://github.com/rainlab/translate-plugin/pull/586

Instead of creating a separate method that has a different behavior to the original implementation, I managed to fix ``transWhere`` itself by utilizing 2 queries instead of the slow orWhere clause. This implementation also preserves the fallback functionality.

I made some tests on my project (originally discussed here: https://github.com/rainlab/translate-plugin/issues/581) - now I have 2 queries that execute in ~ 5 ms instead of a single 60 second query. The improvement is significant.

Comments are welcome.

### DIFF
--- a/behaviors/TranslatableModel.php
+++ b/behaviors/TranslatableModel.php
@@ -46,6 +46,8 @@ class TranslatableModel extends TranslatableBehavior
             $locale = $this->translatableContext;
         }
 
+        // Separate query into two separate queries for improved performance
+        // @see https://github.com/rainlab/translate-plugin/pull/623
         $translateIndexes = DB::table('rainlab_translate_indexes')
             ->where('rainlab_translate_indexes.model_type', '=', $this->getClass())
             ->where('rainlab_translate_indexes.locale', '=', $locale)
@@ -53,7 +55,7 @@ class TranslatableModel extends TranslatableBehavior
             ->where('rainlab_translate_indexes.value', $operator, $value)
             ->pluck('model_id');
 
-        if($translateIndexes->count()) {
+        if ($translateIndexes->count()) {
             $query->whereIn($this->model->getQualifiedKeyName(), $translateIndexes);
         } else {
             $query->where($index, $operator, $value);
--- a/behaviors/TranslatableModel.php
+++ b/behaviors/TranslatableModel.php
@@ -48,7 +48,7 @@ class TranslatableModel extends TranslatableBehavior
 
         // Separate query into two separate queries for improved performance
         // @see https://github.com/rainlab/translate-plugin/pull/623
-        $translateIndexes = DB::table('rainlab_translate_indexes')
+        $translateIndexes = Db::table('rainlab_translate_indexes')
             ->where('rainlab_translate_indexes.model_type', '=', $this->getClass())
             ->where('rainlab_translate_indexes.locale', '=', $locale)
             ->where('rainlab_translate_indexes.item', $index)
--- a/behaviors/TranslatableModel.php
+++ b/behaviors/TranslatableModel.php
@@ -42,19 +42,22 @@ class TranslatableModel extends TranslatableBehavior
      */
     public function scopeTransWhere($query, $index, $value, $locale = null, $operator = '=')
     {
-        $query->select($this->model->getTable().'.*');
+        if (!$locale) {
+            $locale = $this->translatableContext;
+        }
 
-        $query->where(function($q) use ($index, $value, $operator) {
-            $q->where($this->model->getTable().'.'.$index, $operator, $value);
-            $q->orWhere(function($q) use ($index, $value, $operator) {
-                $q
-                    ->where('rainlab_translate_indexes.item', $index)
-                    ->where('rainlab_translate_indexes.value', $operator, $value)
-                ;
-            });
-        });
+        $translateIndexes = DB::table('rainlab_translate_indexes')
+            ->where('rainlab_translate_indexes.model_type', '=', $this->getClass())
+            ->where('rainlab_translate_indexes.locale', '=', $locale)
+            ->where('rainlab_translate_indexes.item', $index)
+            ->where('rainlab_translate_indexes.value', $operator, $value)
+            ->pluck('model_id');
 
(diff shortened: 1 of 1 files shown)


mjauvin:
I really like this implementation @acasar !

mjauvin:
From initial tests I have been doing, it behaves the way I would expect it to.

Can you guys @mariavilaro @mterman @eoler @aurelien-roy test this to see if you get the same performance improvements as @acasar and report back with numbers?

Thanks.

behaviors/TranslatableModel.php
```diff
@@ -42,19 +42,22 @@ public function __construct($model)
...
+            ->where('rainlab_translate_indexes.locale', '=', $locale)
+            ->where('rainlab_translate_indexes.item', $index)
+            ->where('rainlab_translate_indexes.value', $operator, $value)
+            ->pluck('model_id');
+
+        if($translateIndexes->count()) {
+            $query->whereIn($this->model->getQualifiedKeyName(), $translateIndexes);
+        } else {
+            $query->where($index, $operator, $value);
+        }
```
LukeTowers, line 60:
```suggestion
        // Separate query into two separate queries for improved performance
        // @see https://github.com/rainlab/translate-plugin/pull/623
        $translateIndexes = DB::table('rainlab_translate_indexes')
            ->where('rainlab_translate_indexes.model_type', '=', $this->getClass())
            ->where('rainlab_translate_indexes.locale', '=', $locale)
            ->where('rainlab_translate_indexes.item', $index)
            ->where('rainlab_translate_indexes.value', $operator, $value)
            ->pluck('model_id');

        if ($translateIndexes->count()) {
            $query->whereIn($this->model->getQualifiedKeyName(), $translateIndexes);
        } else {
            $query->where($index, $operator, $value);
        }
```

behaviors/TranslatableModel.php
```diff
@@ -42,19 +42,24 @@ public function __construct($model)
...
-            });
-        });
+        if (!$locale) {
+            $locale = $this->translatableContext;
+        }
 
-        $this->joinTranslateIndexesTable($query, $locale);
+        // Separate query into two separate queries for improved performance
+        // @see https://github.com/rainlab/translate-plugin/pull/623
+        $translateIndexes = DB::table('rainlab_translate_indexes')
```
mjauvin, line 51:
```suggestion
        $translateIndexes = Db::table('rainlab_translate_indexes')
```

mjauvin:
I did some tests, created 50,000 records for one of my models, and added a translation for those 50,000 records;

- Without this PR, it took 3.39s to complete a transWhere().
- With this PR, it took 54ms

So I don't have the kind of improvement I heard of, but it's still a significant (62X) improvement.

Good work!

LukeTowers:
@mjauvin I'm happy with this if you are. Great work @acasar!

Samuell1:
Hey, not sure why but this version of transWhere broke searching for slug for random slugs when i rolled it back to older version with leftJoin it works as expected.

daftspunk:
It might be worthwhile to keep the old implementation alongside the new one. What to call it?

acasar:
I think what @iboved is referring to is to have an option to add a whereTrans clause without a default fallback. Bringing back the original implementation would not solve that because it worked the same way - if translated slug was not found, it checked the default slug with an orWhere clause.

Another parameter would need to be introduced to enable that behavior. In few cases where I needed that I created my own scope and joined translated table manually. But it may be nice to have this option by default.

daftspunk:
Got it. Something added for this here: fdec343b8540cb7f90584169747df5fe9308f158

iboved:
Thanks, guys, for the quick response and solution!
Did I understand correctly that the `scopeTransWhereNoFallback` method works only with **non-default** locales? It searches only in the `rainlab_translate_indexes` table (while the value from the default locale is stored in the original model table). This means that we can not use this method to search value in the **default** locale. For default locale we should use `scopeTransWhere`. Is it correct behavior?

acasar:
Actually you should just use a normal where for the default locale, since you don't want to query translatable data in this case.

```php
$trans = Translator::getInstance();

if($trans->getLocale() === $trans->getDefaultLocale()) {
    $post = Post::where('slug', $slug)->first();
} else {
    $post = Post::transWhereNoFallback('slug', $slug)->first();
}
```

But I agree that this condition should probably be handled within ``transWhereNoFallback`` method. I made a PR with a suggested fix here: https://github.com/rainlab/translate-plugin/pull/685 Although, maybe this need some more discussion.... 

Currently we have the following situation. Imagine we have 2 Posts:

| Locale  | Post 1 (slug) | Post 2 (slug) |
| ------------- | ------------- | ------------- |
| en  | first-post  | second-post |
| sl  | prvi-clanek  | null |

If we use the following where conditions in **en** locale, we get these results:

| Condition (en) | first-post | second-post | prvi-clanek |
| ------------- | ------------- | ------------- | ------------- |
| transWhere | Post 1 | Post 2 | No result |
| transWhereNoFallback | Post 1*  | Post 2*  | No result |

_\* No result without this PR: https://github.com/rainlab/translate-plugin/pull/685_

If we use the following where conditions in **sl** locale, we get these results:

| Condition (sl) | first-post | second-post | prvi-clanek |
| ------------- | ------------- | ------------- | ------------- |
| transWhere | Post 1 | Post 2 | Post 1 |
| transWhereNoFallback | No result  | No result | Post 1 |

But I think what we actually want when we run the query in **sl** locale is this:

| Condition (sl) | first-post | second-post | prvi-clanek |
| ------------- | ------------- | ------------- | ------------- |
| _Ideal behavior_ | No result | Post 2 | Post 1 |

So currently no implementation of ``transWhere`` is perfect. The old implementation worked in the same way as ``transWhere`` works now, so it had the same issue. But I don't know, maybe this is not that critical, since the only "side effect" of using ``transWhere`` in a foreign locale is that it returns a result in two cases - when using the original slug and also when using the translated slug. We never really had any issues with that, since links with the original slugs do not appear anywhere on the site and are therefore not indexed by Google. We also set canonical links to all our pages, which further minimizes the issue.

PR #625, opened 2020-11-17
only extend if it is not already extended

mjauvin:


LukeTowers:
@mjauvin I thought this was discussed in an earlier issue and the determination is that this "fix" is not fixing the underlying issue that triggers it, it's only a band aid that covers it up. @bennothommo do you remember this discussion at all?

mjauvin:
@LukeTowers 
If multiple plugins extend the same model with a behavior, this becomes a nightmare to manage.

It's a bit like trying to access an array element without checking if the key exist in the first place... 

I would personally modify the extendClassWith() method to to this check, but seems this is not the concensus.

LukeTowers:
If other plugins are extending those core models with those behaviours then that's the other plugins fault, only this plugin should be doing that. Please point to an example of another plugin extending those three models with this plugin's behaviours, or at least better explain what exact issue this is solving

mjauvin:
I don't have the reports now, but I've had people reporting problems with their plugins which were already extending those models before this was added to the Translate plugin itself.

I don't see why two plugins can't extend the same class, if it has already been extended, fine, move along. Why does it need to break ?

bennothommo:
@LukeTowers it was a fair while ago, so I don't remember the full discussion, but it does ring a bell.

On one hand, I agree with @LukeTowers in that other plugins *shouldn't* be extending the model with the same behaviour, as it doesn't make a great deal of sense for the behaviour to be used outside of the Translate plugin.

On the other hand, if we have a way of gracefully handling repeated extension calls with the same behaviour - in this case, simply ignoring the second extension call - then I don't have a good answer as to why we shouldn't take advantage of that. If two plugins require a method that is provided by a behaviour, then there's no harm in having the behaviour be implemented irrespective if one or both try to implement it. Both plugins should theoretically be able to work harmoniously with the behaviour.

Unless there's some mechanic of the behaviours that I'm not aware of that would preclude this scenario from working...

PR #649, opened 2021-08-11
Change self::makeCacheKey and self::makeMessageCode to static::

panki3a:
Change calling protected method makeCacheKey and makeMessageCode with static, to enable call override method from custom Message class.
example:
class Message extends \RainLab\Translate\Models\Message
{
    /**
     * @param string $messageId
     * @return string
     */
    protected static function makeMessageCode($messageId)
    {
        return $messageId;
    }
}

### DIFF
--- a/models/Message.php
+++ b/models/Message.php
@@ -109,7 +109,7 @@ class Message extends Model
             return $messageId;
         }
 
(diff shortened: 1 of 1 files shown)


daftspunk:
Thanks!

PR #670, opened 2022-04-07
rainlab_translate_attributes deletion on afterDelete

massimomegistus:
This change gives an opportunity  of doing something before rainlab_translate_attributes are deleted


daftspunk:
Good! Thanks

PR #674, opened 2022-04-19
Fix incorrect docs about content & mail template translation

guus-frenken:
The suffix is supposed to start with `-`, not `.`
https://github.com/rainlab/translate-plugin/blob/master/classes/EventRegistry.php#L394


daftspunk:
Thanks!

Ladylain:
Unfortunately, on v3.1 the suffix "-" doesn't work, it's a "."

PR #683, opened 2022-06-24
Fix for parent theme not getting scanned for messages, except when the child theme configuration throws an exception

marcogrueter:
In a parent/child theme setup, the parent theme nevers gets scanned for messages, except when the child theme supplies an invalid configuration (like non-existant file or syntax error in the YAML file - something that throws an exception).

This PR lets scanThemeConfigForMessagesInternal return false if no configuration value is present,
so the part of ThemeScanner.php that handles parent theme translations can check for false and try the parent theme.

If a translate config key is present in the child, only the child configuration gets used, as intended.

### DIFF
--- a/classes/ThemeScanner.php
+++ b/classes/ThemeScanner.php
@@ -83,7 +83,9 @@ class ThemeScanner
             $parentTheme = $theme->getParentTheme();
 
             try {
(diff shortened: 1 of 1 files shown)


daftspunk:
Makes sense. Thanks

PR #703, opened 2022-10-15
Make importMessages public

acasar:
I believe this is needed if we want to add custom messages to the scanner:

```php
Event::listen('rainlab.translate.themeScanner.afterScan', function(ThemeScanner $scanner) {
    $scanner->importMessages(['Custom message 1', 'Custom message 2']);
});
```


daftspunk:
Ah, yes! Thank you

PR #743, opened 2024-05-24
Fix issue when duplication relation

acasar:
This solves issue number 2) described here: https://github.com/octobercms/october-private/issues/615


daftspunk:
Nice! Thanks

PR #756, opened 2025-03-25
Performance. Change column 'model_id' to int. Change index.

i3bepb:
The current implementation with integer ID models causes inefficient index usage where only the model_type index is employed, leading to excessive row scanning. Suggested solution: modify the model_id data type and implement a compound index.
image

After
image

### DIFF
--- a/updates/update_attributes_change_index_model.php
+++ b/updates/update_attributes_change_index_model.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace Aerotur\Nemo\Updates;
+
+use DB;
+use Log;
+use October\Rain\Database\Updates\Migration;
+use Schema;
+
+class UpdateAttributesCahngeIndexModel extends Migration
+{
+    public function up()
+    {
+        // Check if there are any non-numeric values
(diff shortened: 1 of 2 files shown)


apinard:
I'm wondering what would happen to existing records where model_id is not in a numeric format (e.g., UUID). Would they be truncated?

FlusherDock1:
> I'm wondering what would happen to existing records where model_id is not in a numeric format (e.g., UUID). Would they be truncated?

This is not a typical situation, in that case the developer can modify this column back to varchar and use it as before

i3bepb:
If the SQL query was formed with the value 1772 in quotes, the index would probably be used. It might be worth changing the query generation code.

i3bepb:
> I'm wondering what would happen to existing records where model_id is not in a numeric format (e.g., UUID). Would they be truncated?

Do you have the opportunity to review the EXPLAIN plan for this query with UUID?

daftspunk:
Why closed? Looks like a good change, akin to what we did in the `system_files` table.

We solved string values by renaming the column first, introducing the new integer column and then copying the values from old column to new column.

